PR #42, opened 2016-01-10
FIX: renumberSerial activated

j0kaso:
renumberSerial option previously not activated


arose:
Looks good, thanks!

Eventually we should have unit tests for such things

j0kaso:
for the next time, do you also want updated js files produced by build
included?
Am 12.01.2016 03:07 schrieb "Alexander Rose" <notifications@github.com>:

arose:
> for the next time, do you also want updated js files produced by build included?

no build files please

PR #102, opened 2016-05-03
CHANGED PDB writer.js: position of atomname

j0kaso:
previous position of atomname for pdb writer caused problems in many tools e.g. Schrodinger/PyMol Widgets... now similar to RCSB PDB provided formats.

### DIFF
--- a/js/ngl/writer.js
+++ b/js/ngl/writer.js
@@ -85,7 +85,7 @@ NGL.PdbWriter = function( structure, params ){
                 // Alignment of one-letter atom name such as C starts at column 14,
                 // while two-letter atom name such as FE starts at column 13.
                 var atomname = a.atomname;
-                if( atomname.length === 1 ) atomname = " " + atomname;
+                atomname = " " + atomname;
 
                 records.push( sprintf(
                     formatString,


arose:
Can you give an example?

js/ngl/writer.js
```diff
@@ -85,7 +85,7 @@ NGL.PdbWriter = function( structure, params ){
                 // Alignment of one-letter atom name such as C starts at column 14,
                 // while two-letter atom name such as FE starts at column 13.
                 var atomname = a.atomname;
-                if( atomname.length === 1 ) atomname = " " + atomname;
+                atomname = " " + atomname;
```
arose: What happens with atomnames of 4 characters? The field is then 5 characters long but only 4 are allowed. Also, the comment does not match the code.

j0kaso:
an example would be: import e.g. 3sn6 and download it - compare then the first 6 atom name positions
ngl-download:
ATOM      1  N   THR A   9      62.538 -14.787  23.643  1.00171.15           N
ATOM      2 CA   THR A   9      62.599 -16.149  23.127  1.00171.98           C
ATOM      3  C   THR A   9      62.653 -17.161  24.270  1.00166.96           C
ATOM      4  O   THR A   9      61.992 -16.993  25.295  1.00166.22           O
ATOM      5 CB   THR A   9      63.831 -16.342  22.215  1.00174.37           C
ATOM      6 OG1  THR A   9      65.024 -16.274  23.005  1.00172.10           O
ATOM      7 CG2  THR A   9      63.876 -15.265  21.146  1.00175.31           C
RCSB PDB download:
ATOM      1  N   THR A   9      62.538 -14.787  23.643  1.00171.15           N  
ATOM      2  CA  THR A   9      62.599 -16.149  23.127  1.00171.98           C  
ATOM      3  C   THR A   9      62.653 -17.161  24.270  1.00166.96           C  
ATOM      4  O   THR A   9      61.992 -16.993  25.295  1.00166.22           O  
ATOM      5  CB  THR A   9      63.831 -16.342  22.215  1.00174.37           C  
ATOM      6  OG1 THR A   9      65.024 -16.274  23.005  1.00172.10           O  
ATOM      7  CG2 THR A   9      63.876 -15.265  21.146  1.00175.31           C  

I know so far that Schrödinger can not recognize the sequence with this atom name position. Also SL2 has problems with that (there, the atom name positions are oriented according to the pdb - see link below).

I did not know, that atom names with 5 letters are supported within the pdb file format. There are my known information:
> 13 - 16        Atom          name         Atom name.
[(PDB File Format)](http://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#ATOM)
Following this guidelines, your starting point for the atom names is shifted by one position (therefore I inserted the space).

If you prefer, I can also simply claim this as a bug.

arose:
> I did not know, that atom names with 5 letters are supported within the pdb file format.

They are not, that is my point, your change may thus break.

Could you please format the pdb lines as code so they are fixed width it is hard see spaces. Is is output before or after your change?

j0kaso:
changed the comment. I just loaded the structure from the RCSB pdb (via PDB import) and then downloaded it. nothing more

js/ngl/writer.js
j0kaso: In my opinion, you chose the wrong starting position for the atom names. If you start all from the same point (removing the if cause), one space is missing according to the RCSB pdb file format.

arose:
Ok, I see. Here is a [real example](http://www.rcsb.org/pdb/explore/explore.do?structureId=1EJG) showing the alignment for atomnames of all lengths. It would be great if you could change the PR to reproduce that. Thanks so much!
```
ATOM    146  CG2 VAL A   8       3.635   1.036  13.955  0.55  4.82           C  
ATOM    148  H   VAL A   8       6.374   2.231  14.084  0.55  2.27           H  
ATOM    150  HA  VAL A   8       4.948   1.364  11.845  0.55  2.65           H   
ATOM    152  HB  VAL A   8       4.204   3.077  14.197  0.55  4.16           H   
ATOM    154 HG11 VAL A   8       2.862   3.979  12.286  0.55  6.22           H  
```

PR #111, opened 2016-05-09
Fix bug in mmCIF parser

sbliven:
Loading cif files with spacegroup info failed due to typo

### DIFF
--- a/js/ngl/parser.js
+++ b/js/ngl/parser.js
@@ -2782,8 +2782,8 @@ NGL.CifParser.prototype = NGL.createObject(
 
             var symmetry = cif.symmetry;
 
-            var sGroup = symmetry[ "space_group_name_H-M" ];
-            if( sGroup[0] === sGroup[ sGroup.length-1 ] &&
+            var sGroup = symmetry[ "space_group_name_H_M" ];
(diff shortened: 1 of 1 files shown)


js/ngl/parser.js
```diff
@@ -2782,8 +2782,8 @@ NGL.CifParser.prototype = NGL.createObject(
 
             var symmetry = cif.symmetry;
 
-            var sGroup = symmetry[ "space_group_name_H-M" ];
-            if( sGroup[0] === sGroup[ sGroup.length-1 ] &&
+            var sGroup = symmetry[ "space_group_name_H_M" ];
```
josemduarte: The field in mmCIF has a hyphen between H and M i.e. `space_group_H-M`. Does it work with underscore?

arose:
Hmm, the mmcif dict says it is a [hyphen]( http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v40.dic/Items/_symmetry.space_group_name_H-M.html). I am ok with adding the underscore variant if there are files out there needing it.

hainm:
> I am ok with adding the underscore variant if there are files out there needing it.

It's better to force people to follow standard.

sbliven:
The cif file i'm using was generated by biojava, so if there are errors we should fix them in biojava not ngl

PR #115, opened 2016-05-11
Orthographic

arose:
TODO
- [x] panning with orthographic camera not good
- [x] screenshots broken (for orthographic camera)
- [x] fix hyperball impostor (just noise) -> will likely be removed (#118)
- [x] fix sphere impostor (bounding box too small)
- [x] fix cylinder impostor (flickering)


hainm:
fantastic

Issue #138, opened 2016-06-02
Double and triple bond depiction


fredludlow:
Would be great to have this! (Related: #133).

Very happy to have a go at this, though would definitely need a few pointers as to how to best do it and perhaps not the easiest first foray into the code... My thoughts so far, (thinking out loud so apologies if nonsense :) ).

* Presumably only want to do this for a few representations (line, licorice and ball+stick mode).
* I'm guessing some kind of offset line or (slimmer) cylinder would be a good way to represent higher order bonds?
* Do the extra lines/cylinders live in the same `LineBuffer`/`CylinderBuffer` as the current bonds... 
 * This could break the alignment of the `LineBuffer` with internal `bondSet`, so might be problematic?
 * ...or create multiple buffers (main bond, secondary bond, tertiary bond)?
* Sensible positioning of the extra lines/cylinders requires placing in plane plus (basic) ring detection to ensure they're placed on the inside of rings
* Should calculation of the position of extra lines/cylinders happen in the structure object or representation? (I guess it's a representation property?).
* Would want to maintain option to turn off

arose:
> Presumably only want to do this for a few representations (line, licorice and ball+stick mode).

Sure, eventually. In the beginning it is probably good to just create an extra representation to simplify things. When creating new representations I like to start by just creating `BufferRepresentation`s.

```
var sphereBuffer = new NGL.SphereBuffer(...);
component.addBufferRepresentation();
```

This avoids needing to care about all sorts of details when creating a new representation.

> I'm guessing some kind of offset line or (slimmer) cylinder would be a good way to represent higher order bonds?

Yes, just smaller cylinders, shifted in a plane.

> Do the extra lines/cylinders live in the same LineBuffer/CylinderBuffer as the current bonds...
> - This could break the alignment of the LineBuffer with internal bondSet, so might be problematic?
> - ...or create multiple buffers (main bond, secondary bond, tertiary bond)?

Create extra buffers

Single bonds
`var cylinderBuffer1 = new NGL.CylinderBuffer(...);`

Double bonds
`var cylinderBuffer1 = new NGL.CylinderBuffer(...);`
`var cylinderBuffer2 = new NGL.CylinderBuffer(...);`

Triple bonds
`var cylinderBuffer1 = new NGL.CylinderBuffer(...);`
`var cylinderBuffer2 = new NGL.CylinderBuffer(...);`
`var cylinderBuffer3 = new NGL.CylinderBuffer(...);`

Quadruple bonds
...

[...]

PR #148, opened 2016-06-17
Working bond-orders!

fredludlow:
#138 
This is the merged version - I've also included data and code for an example called bondOrders. This loads the ligand from PDB 4umt (in sdf format), and displays a licorice representation (ball+stick also works for obvious reasons). 

Apart from general optimization, I guess other things to do are:
- [ ] LineRepresentation 
- [ ] Asymmetric offsets 
- [ ] Quadruple bonds? (rare)
- [ ] Better colinear fallback? (again, rare)

Let me know what I can do, and thanks so much for doing > 95% of this!

Fred


arose:
This awesome, thanks so much!

> general optimization

Will do

>  LineRepresentation

I'll have a look at that

> Asymmetric offsets

Feel free to introduce a parameter for that

> Quadruple bonds? (rare)

Do you have an example file?

> Better colinear fallback? (again, rare)

what is the current fallback?

arose:
Hi, I moved the calculation of the shift vector to the bond proxy class, so that the residue type class does not has to handle actual data of specific atoms.

AtomProxy objects have .x/.y/.z properties so there is often no need to copy the data into a Vector3 object.

Also fixed a bug in the bitset class that that lead to invisible spheres at the origin and broke the stage.centerView().

arose:
I have been looking at a few other examples and it appears the reference atoms are not always correctly chosen. 

- http://files.rcsb.org/ligands/download/2NC_model.sdf
- http://mmtf.rcsb.org/v0.2/full/1crn.mmtf

This is with your PR version and the latest version with my changes.

fredludlow:
>> Asymmetric offsets

>Feel free to introduce a parameter for that

Will take a look...

>> Quadruple bonds? (rare)

> Do you have an example file?

I don't know of any PDB examples but I'm also not entirely sure how best to find them. You get things like Cr-Cr and Mo-Mo quadruple bonds in small molecule X-ray structures. (Probably one for when it gets requested by someone, though it would be fairly trivial to do...) mmCIF format supports quadruple bonds so could hack a nonsense example together

>>Better colinear fallback? (again, rare)

> what is the current fallback?

Urm, there isn't one, sorry... partly because I didn't do the experiment to see how close to 1 the dot product term needs to be before the precision of the normalized shift vector suffers visibly - though if they are absolutely colinear (not impossible given reduced precision of input coordinates) then you end up trying to normalize (0,0,0)! One algorithm would be simply:
* If ref atom present and not colinear, use it
* If (1,0,0) not colinear, use that (ignoring other connected atoms that might be more appropriate: I think most of these examples are going to be alkynes (other atom no better) or "complicated" heavier elements, where all bets are off anyway). 
* If (0,1,0) not colinear, use that. 


> Hi, I moved the calculation of the shift vector to the bond proxy class, so that the residue type class does not has to handle actual data of specific atoms.
> AtomProxy objects have .x/.y/.z properties so there is often no need to copy the data into a Vector3 object.
> Also fixed a bug in the bitset class that that lead to invisible spheres at the origin and broke the stage.centerView(). 

Ah, I wondered what was going on there. (And I'm enjoying my lessons in performant JS while looking through the code!)

> I have been looking at a few other examples and it appears the reference atoms are not always correctly chosen.

Embarrassing - I'll have a look...

fredludlow:
Turns out there was all types of wrong going on in the stuff I'd added to residue-type (I changed strategy halfway though and, well, code-coverage/unit-tests would have prevented this... one day I'll learn). Two other minor bugs squashed too: getBondIndex didn't reset idx2 in the outer loop so skipped a few possibilities, and the residueAtomOffset wasn't getting applied to the result of getReferenceAtomIndex.

Those two files look okay for me now (unless I'm missing something else)

arose:
> code-coverage/unit-tests would have prevented this...

I started setting up unit-tests but it is not quite usable (a bit complicated due to using es6 modules)

> Two other minor bugs squashed too

Thanks for spotting those!

> Those two files look okay for me now (unless I'm missing something else)

Can you do a PR?

PR #155, opened 2016-07-01
StructureRepresentation.selection referenced before definition

fredludlow:
Hi again,

Just a quick one - If I load a structure then add a cartoon representation I get an error because StructureView hasn't got a selection attribute defined yet. I think this is down to the order attributes are defined in StructureRepresentation. This PR fixes it for me (though that's possibly more by luck than judgement!)

Cheers,

Fred

### DIFF
--- a/src/representation/structure-representation.js
+++ b/src/representation/structure-representation.js
@@ -42,6 +42,12 @@ function StructureRepresentation( structure, viewer, params ){
     this.structure = structure;
 
     /**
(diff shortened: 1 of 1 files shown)


arose:
You are exactly right, I already have that fixed locally, will upload later. Thanks

arose:
done 636964ca4c2e23daf3c9f4de3018ae7501d0fdd8

Issue #179, opened 2016-08-10
Detect rings in groups/residues


arose:
Up to 6 (or 7?) ring members.

Atoms can belong to multiple rings. Data can be stored in ResidueType instances.

arose:
@fredludlow how far are we from supporting that?

fredludlow:
Was looking at this when originally thinking about the reference atoms for bond shifts stuff. I guess "find all the rings" is not quite the same algorithm I'm using for "find the smallest ring containing this bond" (just DFS), I can have a look at what RDKit or another permissively-licensed chemoinformatics toolkit does. 

I was looking at refactoring the reference atom stuff and I think these various features would benefit from a really quick way to traverse the molecule - BondHash nearly does that, but it gives you a bond Index and have to work out what the "other" atom is (sorry, late here and not explaining well). That could be on-the-fly: BondProxy.getOtherAtom(atomproxy) or AtomProxy.getBondedAtom(bondproxy); or could have another array in BondHash of "otherAtomIndex" - (increases storage, most of which never accessed...)

arose:
> Was looking at this when originally thinking about the reference atoms for bond shifts stuff. I guess "find all the rings" is not quite the same algorithm I'm using for "find the smallest ring containing this bond" (just DFS), I can have a look at what RDKit or another permissively-licensed chemoinformatics toolkit does.

Actually I currently only need a flag per atom indication if it is member of a ring or not. So I would welcome that as a first step :-)

> That could be on-the-fly: BondProxy.getOtherAtom(atomproxy) or AtomProxy.getBondedAtom(bondproxy); or could have another array in BondHash of "otherAtomIndex" - (increases storage, most of which never accessed...)

Or `AtomProxy.eachBondedAtom( callback )`?

fredludlow:
> Actually I currently only need a flag per atom indication if it is member of a ring or not. So I would welcome that as a first step :-)

In that case this should do it: RDKit's [fastFindRings...](https://github.com/rdkit/rdkit/blob/master/Code/GraphMol/FindRings.cpp#L1231)That method doesn't find all rings but does find all ring atoms, if that makes sense.

Actually, storing the rings also simplifies the reference atom stuff. 

> Or AtomProxy.eachBondedAtom( callback )?

Yup. why don't I do that and ResidueType.assignRings and if that's any good I'll use them for the multiple bond stuff (I've got a first attempt at the asymmetric offset bonds thing, will get it out tomorrow)

Issue #152, opened 2016-06-23
Asymmetric offsets for double bonds


arose:
To draw extra cylinders/lines inside a ring

![download](https://cloud.githubusercontent.com/assets/272250/16309669/6d9bd668-391e-11e6-843c-e79b7d79ff34.png)

fredludlow:
I've got an initial commit for this here: fredludlow/ngl@11faf5d3295ca3deb314803f87e5d180079cf313

Looks nice for lines, looks okay for licorice (the cylinders are uncapped - might need to place extra spheres at the ends to round them off?) and looks a bit sketchy for ball+stick mode.

![screenshot 3](https://cloud.githubusercontent.com/assets/900535/17624759/ea549a68-609d-11e6-899b-255f97cec3de.png)
![screenshot 4](https://cloud.githubusercontent.com/assets/900535/17624760/ea599e3c-609d-11e6-90a1-300250b49d75.png)
![screenshot 5](https://cloud.githubusercontent.com/assets/900535/17624761/ea5cf474-609d-11e6-8eb5-be1f18d84725.png)

arose:
Looks great!

Does the amount of "shortening" need to be adjusted for different values of `bondSpacing`?

> the cylinders are uncapped - might need to place extra spheres at the ends to round them off?

Yeah would be nice, but can be done in the future. Either with extra spheres or rounded caps (https://github.com/arose/ngl/issues/187). In both cases it requires some refactoring as the placement of the spheres and the "identity" of the shifted bonds is only known within the `getBondData` method.

> looks a bit sketchy for ball+stick mode.

I guess that is okay. One could change the `aspectRatio` param for bigger spheres.

fredludlow:
Thanks! 

> Does the amount of "shortening" need to be adjusted for different values of bondSpacing? 

Probably - actually the parameters need a bit of a think through (I'm currently abusing the `bondSpacing` parameter because I just wanted a way to play with relative radii in the UI easily without wiring up another parameter).

I think for Balll+Stick the current (symmetric) rendering looks better anyway (it looks a bit more like the plastic molecular modelling kits you can buy) so maybe for that mode at least the current (two smaller symmetric cylinder) mode should be default. 

[...]

PR #186, opened 2016-08-12
Added ring detection and flags

fredludlow:
As discussed in #179. See the comments in the docstring and example "ringFlags" in the webapp, this is a fast but non-exhaustive method for finding rings, using the RDKit fastFindRings method. A couple of points/possible issues:

* I access bondHash directly from structure-utils, this might be a bad thing? (breaks encapsulation). I can refactor to use the eachBondedAtom callback and similar, but that creates a ton of callbacks and I was slightly tying myself up to make sure I wasn't reusing proxy objects at multiple depths of recursion.

* For really pathological structures (single residue with 10s of thousands of atoms) you might get a stack overflow...


arose:
If ring detection is per residue (ResidueType), would it not be sufficient to use  `ResidueType.bonds.atomIndices1/2`? Maybe create something similar to BondHash to conveniently get bonded atoms? Maybe it can be combined with `propagateSearch` and `ResidueType.assignBondReferenceAtomIndices`.

As the rings object is not complete, is worth keeping/returning it?

> I access bondHash directly from structure-utils, this might be a bad thing? (breaks encapsulation). I can refactor to use the eachBondedAtom callback and similar, but that creates a ton of callbacks and I was slightly tying myself up to make sure I wasn't reusing proxy objects at multiple depths of recursion.

I would consider `bondHash` public, it is just not convenient but  totally okay to use.

The method would be used only once per ResidueType so efficiency is not an issue, I think.

> For really pathological structures (single residue with 10s of thousands of atoms) you might get a stack overflow...

We could test for residues with >1000 atoms, write a warning and return.

fredludlow:
> If ring detection is per residue (ResidueType), would it not be sufficient to use ResidueType.bonds.atomIndices1/2? Maybe create something similar to BondHash to conveniently get bonded atoms? Maybe it can be combined with propagateSearch and ResidueType.assignBondReferenceAtomIndices.

What I had been thinking was: 
* Ring detection and sets of rings as in this PR. 
* Refactor `assignBondReferenceAtomIndices` to use the BondHash object for finding neighbour atoms
* Remove `propogateSearch`: reference atoms for non-ring atoms are just "any bonded atom" while reference atoms for ring atoms are "any other atom in the same ring" - this might give odd results for some big fused ring systems, but I thought they could be fixed by improving the set of rings generated by this code (though I haven't really looked at how to do that yet).

To avoid using `BondHash` and just use the data in `ResidueType`:

* A new 'ResidueType.getBondGraph/Hash' method - uses `ResidueType.bonds.atomIndices1/2`  to return some kind of traversable object (`BondHash`-like, or simple nested JS objects) as you suggested above. 
* `getRings( r )` is a method on `ResidueType` as in this PR but uses `getBondGraph/Hash` (move code from `calculateResidueRings` to a `ResidueType` method?)
  * If necessary a better method of ring detection so that we can pick the "best" ring for an atom (in terms of where you render double bonds etc)
* `assignReferenceAtomIndices` uses both `getRings` and `getBondGraph/Hash` but the logic is otherwise more or less the same.

> As the rings object is not complete, is worth keeping/returning it?

Only for identifying reference atoms quickly (I think it should be sufficient for that in most cases, should probably test it on a heme!)


Let me know what you think.

arose:
> To avoid using BondHash and just use the data in ResidueType:

I would chose this path as it allows re-use of the calculations for all residues of the ResidueType

> move code from calculateResidueRings to a ResidueType method?

sounds good

fredludlow:
This is still WIP - I've refactored `getBondGraph` and associated stuff to method on `ResidueType` and reference atom detection is now ring aware. I grabbed my updated offset vector code too and for me this is working (i.e. bonds are offset and inside rings but now using just the one ring search - which is also exposed for general use). I'm at a conference for the rest of the week but hopefully will get a chance to tidy this up so there are no conflicts at least.

arose:
That sounds like progress, thanks.

fredludlow:
Apologies for the monolithic PR. This should now have both the ring detection in a (fairly) sane way plus the improvements for double bonds discussed in #152

arose:
awesome!

Issue #205, opened 2016-09-21
AV-style surfaces


fredludlow:
So I had a go at porting the aforementioned AstexViewer surface code:

The current method (EDT - the isoquinoline looks a bit weird to me):
![screenshot 12](https://cloud.githubusercontent.com/assets/900535/18719361/02562e88-801f-11e6-82e4-fb7a749649aa.png)

New style (looks a bit more regular):
![screenshot 11](https://cloud.githubusercontent.com/assets/900535/18719407/2f262710-801f-11e6-932c-e6b5ed5beb34.png)

Also works on large surfaces:
![screenshot 13](https://cloud.githubusercontent.com/assets/900535/18719463/5e03a7ec-801f-11e6-934c-e4bd1ea3dfcb.png)

You can play with it [here](http://fredludlow.com/ngldev/examples/webapp.html). Create a surface and turn off useworker (or load the avSurf example) - (basically I set it so that if you generate without using a worker it calls AVSurface and if you use a worker it uses EDTSurface). 

In terms of implementation the `AVSurface` class is a drop-in replacement for the EDTSurface class. 

Code is here:
https://github.com/fredludlow/ngl/tree/av-style-surface

I'll do a bit of tidying etc before making a PR (have to leave the office in like, 2 minutes....)

arose:
This is absolutely amazing!

> New style (looks a bit more regular)

That's a gross understatement ;-)

Let me know if you need help making it work in a Worker.

In terms of integrating it I would suggest to add another type "av" and in `MolecularSurface.getSurface` use `AVSurface` when `params.type` is "av".

If we can make AVSurface a bit faster it should replace EDTSurface.

fredludlow:
Thanks! Though most of the credit should go to Mike for coming up with the method and you for having a such an easy to extend codebase! 

I'll see if I can get it going in a worker and do as you suggest with the "av" type etc. Any worker-specific gotchas I need to look out for? (I was just going to copy the way EDTSuface does it)

I dropped the number of torroidal samples to 20 and I actually get comparable speed between EDT and AV on chrome and IE on my windows box, you can also reduce the scaleFactor to about 1.5-1.6 in AV mode and it still looks okay. However, the difference on Chrome running on Android is still big - EDT seems to win by a factor of 2.5. I don't really understand this.

fredludlow:
[...]

PR #209, opened 2016-09-27
Av style surface

fredludlow:
Also see #205 

I've refactored this as one big closure now and it's much quicker!
Updated my version here:
http://fredludlow.com/ngldev/examples/webapp.html

On Chrome and IE11 on Windows this is comparable in speed to EDT at the same grid resolution. Firefox on Android gives similar speeds. For Chrome on Android it's still a factor of 2 or so slower... obviously...?! ;) 

This is probably a reasonable point to merge if you want to (and do please hammer the code into a better style if there's anything horrible looking in there* :)). There might be some performance gains to be had in projectTorus but it's no longer dominating the CPU profile.

*As an aside - what editor do you use? And can I copy your config file for it?


arose:
Thanks, this is nice!

The scaling behavior seems to be worse than EDTsurf.
- 1AON: 23363.472ms (edt ms) vs 63317.274ms (av).
- 4V5A: 37093.969ms (edt ms) vs 75115.215ms (av).

The av surface for 4V5A also is incomplete, see below

screenshot 33

> *As an aside - what editor do you use? And can I copy your config file for it?

Just Sublime 3, with trailing spaces removal and spaces for indentation. I am thinking of using https://github.com/feross/standard to never think about code style again.

fredludlow:
Hmm, will investigate...

fredludlow:
I've not had a look at 4V58 yet, but for 1AON the profiler suggests SpatialHash.within is taking up most of the time - not sure why this would be...? It creates a new plain old JS array each time, maybe I can pass in a reusable array - I'll have a bit more of a hack (so far I've only managed to make it slower).

arose:
Maybe a `eachWithin( callback )` method for spatial hash to avoid making temp arrays?

fredludlow:
So I've improved the performance a bit, will tidy up and make another PR

I don't really understand why the "within" function dominates the profile for large systems - I suspect it's either something to do with pulling the bucket indices etc in and out of scope and cpu cache sizes and some other hand-waving explanations?? Or else allocating deallocating lots of small neighbour arrays (or some other form of handwaving..) :)

Anyway:  I made a version of SpatialHash that uses a grid size optimized for the problem (2.01 * maxRadius) - that means it has to use rounding rather than bitshifting but essentially the same code. 
I also added a withinRadii method which is aware of the atom radii and gives you a more precisely filtered list of neighbours - this saves a few million unnecessary checks later on. 

This is now faster than EDT ms on my box for 1aon and smaller examples (not quite as fast as EDT vws though).

On 4v5a I get out of memory errors with both methods unless I put the scaleFactor down to about 1.0.  EDT looks awful at that scaleFactor and also throws out-of-memory at higher scaleFactors... 

(The out of memory errors actually occur after building the surface - it's during allocation of the colour buffer for the surface triangles..)

I still haven't dug into why the surface colouring isn't working. 

Anyway - like I say, I'll make a PR and anxiously await a report of whether it's any better on your box :)

PR #210, opened 2016-09-29
Improved performance of AVSurface (added AVHash class)

fredludlow:
Newer, faster version, see #209 and #205

### DIFF
--- a/src/surface/av-surface.js
+++ b/src/surface/av-surface.js
@@ -11,8 +11,151 @@ import { uniformArray } from "../math/array-utils.js";
 import {
     computeBoundingBox, v3multiplyScalar, v3cross, v3normalize
 } from "../math/vector-utils.js";
-import SpatialHash from "../geometry/spatial-hash.js";
 import { defaults } from "../utils.js";
+
+
+/* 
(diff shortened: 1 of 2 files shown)


arose:
Awesome, that is much faster indeed :-)

I think the maxRadius for AvHash needs to take the probeRadius into account. At least the surface looks wrong for larger probe radii.

4v5a looks good on my desktop machine, see below for a scaleFactor of 2.0

screenshot 34

fredludlow:
I need a new machine.

Yeah, I'd noticed some weirdness at high probe sizes. The `maxRadius` parameter is already calculated using the extended atomic radius max(rAtom + rProbe) but it needs to be a bit more I guess - perhaps `max(rAtom + rProbe) + (ng/scaleFactor)`.

Unfortunately I'm travelling for the next two weeks without a laptop (a frightening proposition) but will put it on the to-do list for when I'm back.

arose:
I'll have a look at it, thanks.

> Unfortunately I'm travelling for the next two weeks without a laptop (a frightening proposition) ...

Have fun!

PR #219, opened 2016-10-19
Fixes atom-based colouring of 'av' surfaces

fredludlow:
Small change to marching cubes - Also updated the "avSurf" example as demo.

The marching cubes implementation always generates integer `mu` for EDT Surfaces, so safe to use as an index, but with AV Surfaces it was creating non-integer indexes into `atomindex` array.


arose:
Lol, I feel quite stupid now... nice find!

PR #232, opened 2016-11-08
Indexing bug(s) in AVSurface

fredludlow:
Oops, somewhat embarrassing bug in surface generation - for example: 3pqr, av-style surface in current master renders about half of the surface correctly and the other half in a very broken way. I was (consistently) miscalculating cell indexes in AVHash and quite often they'd fall outside the relevant array. I think this fixes it, but you might want to test a few of your other favourite structures.


arose:
thank you!

arose:
looks much better indeed :-)

PR #235, opened 2016-11-15
Orthographic near clipping

fredludlow:
I get some odd clipping behaviour in orthographic camera mode. Basically, as you zoom in very closely to an object it clips out of view - my expectation (which is possibly wrong, of course!) is that it should be possible to have a completely unclipped orthographic projection (-Inf and +Inf near and far clipping planes).

e.g. Load the webapp and the 1blu example, switch to orthographic and zoom in (and in and in) and stuff starts clipping out of view. 

I've got a checked out version [here](http://fredludlow.com/ngl/orthoclip/examples/webapp.html) which behaves how I'd expect it to. It also makes `clipDist` redundant for orthographic mode (I'm not quite sure what `clipDist` does normally...).

It's possible that I'm missing something obvious and that what I've done is a bad idea... feedback very welcome!

### DIFF
--- a/src/viewer/viewer.js
+++ b/src/viewer/viewer.js
@@ -1133,16 +1133,24 @@ function Viewer( eid ){
 
         var nearFactor = ( 50 - p.clipNear ) / 50;
         var farFactor = - ( 50 - p.clipFar ) / 50;
-        camera.near = Math.max( 0.1, p.clipDist, cDist - ( bRadius * nearFactor ) );
-        camera.far = Math.max( 1, cDist + ( bRadius * farFactor ) );
-
-        // fog
 
         var fogNearFactor = ( 50 - p.fogNear ) / 50;
         var fogFarFactor = - ( 50 - p.fogFar ) / 50;
+
+        var near = cDist - ( bRadius * nearFactor );
+        var fogNear = cDist - ( bRadius * fogNearFactor );
+
+        if( camera.type === "PerspectiveCamera" ){
+            near = Math.max( p.clipDist, 0.1, near );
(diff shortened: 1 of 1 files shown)


arose:
I see your point and agree that this wrong behavior, thanks!

However, I still want to have the clipNear functionality, which is to clip things that are too close to the camera. Started experimenting with how to do it properly for orthographic camera mode.

fredludlow:
My patch should still allow clipNear/Far (and fogNear/Far) it's just clipDist that's ignored (in orthographic mode)

arose:
wrote clipNear instead of clipDist, see my edit

fredludlow:
Ah, that makes sense :-)

arose:
Tweaked it in 82a0f02, basically what you did but having clipDist work when clipNear is 0 for ortho cam.

arose:
@fredludlow does 82a0f02 solves it for you?

fredludlow:
Yes, if I set `clipDist` to zero then `clipNear` and `clipFar` behave as I'd expect them to. 

If you have non-zero `cilpDist` in orthographic mode you still get the "stuff disappearing as I try to zoom in on it" effect. For my use (embedding NGL in another page) it's fine as I can set `clipDist` to zero and ignore it, but it feels slightly unintuitive out-of-the-box (but that may just be down to what I'm used to!)

arose:
> If you have non-zero cilpDist in orthographic mode you still get the "stuff disappearing as I try to zoom in on it" effect.

Yes, but the "disappearing" should be much closer/tighter to the structure. I'll consider changing the default if it feels unintuitive for more people. Thanks for letting me know.

Issue #198, opened 2016-09-08
Sparse surface meshes


fredludlow:
Related to #180 but different enough to warrant its own issue:

I had a bit of a play with making sparser meshes using quads and triangles. [Here](http://fredludlow.com/ngldev/examples/webapp.html)'s a demo! (In the example menu called "contour")

Original (no smoothing, wireframe):
![screenshot 1](https://cloud.githubusercontent.com/assets/900535/18356287/0ca95c10-75e4-11e6-867a-3adb52b393d2.png)

Sparser:
![screenshot](https://cloud.githubusercontent.com/assets/900535/18356294/1394ed6e-75e4-11e6-8c73-e662d0d8a55f.png)

Basic idea is rendering triangle edges from marching cubes but only those edges that fall on a cube face (I think this is actually equivalent to the 2D contouring mentioned before).

The way it's coded at the moment (in my [contour-surface branch](https://github.com/arose/ngl/compare/e072310...fredludlow:contour-surface)) is very much a big hack (I abused the various existing classes to connect the UI and end up returning a LineBuffer rather than a SurfaceBuffer) but the changes in marching-cubes.js outline the meat of it. The current implementation is also quite inefficient (most lines are drawn twice) so plenty of room for optimization. 

By itself this isn't particularly appealing to the eye - I haven't got a way of smoothing these surfaces - but in combination with a better method for generating a continuously valued grid (rather than the binary 1/0 you get with EDT) it could make some prettier (and more accurate) wire-mesh surfaces. 

If you like the look of this I can have a go at turning it from  ugly hack to actual feature (though need some thought/discussion on how to restructure the various classes to do it cleanly).

arose:
I like it! Will think about how this could be cleanly implemented. Maybe instead of `from` and `to` arrays use the `index` array and in the SurfaceBuffer set an option to render as Line (this.line = true;). That way everything else should work.

fredludlow:
So in marching-cubes.js `positionArray` stores triangle vertices, and `indexArray` specifies which ones to use to make the triangles (length of `indexArray` is 3x number of triangles). Do you mean that in line mode `indexArray` would now be 2x number of lines (from and to)? (And then appropriate handling depending on value of SurfaceBuffer.line flag)?

Another thing I'd just ignored was the `atomIndex` array, though I guess that's quite important (to get the `filterSele` functionality)

arose:
>  in line mode indexArray would now be 2x number of lines (from and to)?

Exactly

[...]

PR #236, opened 2016-11-17
Contoured Surfaces

fredludlow:
Hopefully ready now! (see #198)

I was about to send a PR with the `edgeFilter` bits removed and a comment on how there were still occasional holes in the surface, turns out I'd made a big error in how I loop over triangles (see 63c1bb0). This version has hole-free surfaces (I hope) and I've reinstated `edgeFilter` to reduce unnecessary duplicated lines (which might make a small difference on huge surfaces)

### DIFF
--- a/examples/js/examples.js
+++ b/examples/js/examples.js
@@ -1471,19 +1471,19 @@ NGL.ExampleRegistry.addDict( {
             } );
             o.addRepresentation( "surface", {
                 sele: "hetero and (not water)",
-                useWorker: true,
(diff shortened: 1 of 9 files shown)


arose:
Thanks @fredludlow this is great! I could have a look at enabling this for surfaces generated from volumes or do you plan to do that?

fredludlow:
That'd be great if you could! (It occurred to me it should be possible for e.g. electron density maps but I haven't looked at it).

PR #237, opened 2016-11-24
Add some padding when selecting which atoms to use for filtered surfaces

fredludlow:
I get slightly different surfaces in these two cases
1) Create a new surface representation with the filterSele supplied at creation time (green below)
2) Create a new surface representation with no filterSele and subsequently add it (red/brown below)
surface-bug
surface-fixed

The green version is wrong I think - it doesn't include enough of the protein so draws protein surface where it isn't - this PR adds a small amount of padding (6.0A) when selecting which atoms to calculate surface on initially and fixes the first case (now looks like the second red/brown one).


fredludlow:
Just bumping this back onto the radar. Can it be merged or should I change or close it?

arose:
looks good. thanks for bringing this back to life :) I might change it but it is definitely better than what was there before.

Issue #255, opened 2017-02-04
CSV autoLoad shows error on Firefox


vvv444:
**Reproduction**
Use autoLoad for some CSV file: `NGL.autoLoad(myCsvUrl, { ext: 'csv' })`

**Problem**
On Firefox the log in developer console shows the following error:
> XML Parsing Error: syntax error Location: ...

Although this does not present any actual problem, it is still annoying.
This happens because if no responseType is specified for the XMLHttpRequest object, Firefox automatically assumes it is XML and attempts to parse it internally.

vvv444:
Related info: http://stackoverflow.com/questions/14399648/reading-text-file-trough-xmlhttprequest-produces-syntax-error-in-text-file-work

arose:
thanks, great find!

Issue #254, opened 2017-02-04
File load error is not reported


vvv444:
**Reproduction scenario**
Name some wrong formatted file with ".mmtf.gz" extension and try to load it.

**Expected behaviour**
NGL.Stage.loadFile() shall raise as exception.

**Actual behaviour**
No exception is thrown. Moreover, there is a non-decremented task left.

**Details**
An exception is raised from MmtfParser._parse(), but this function is called in context originating from 'load' event of XMLHttpRequest() in NetworkStreamer._read() which lacks exception handling.

arose:
thanks!

PR #256, opened 2017-02-04
NetStreamer bug fixes

vvv444:
This solves #254 and #255.


arose:
thanks!

With commit 42cd7de8f0b1a87910c9c55d16941a3f0c39c848 I also started creating a Promise already on the Streamer class which should make error handling more transparent.

PR #257, opened 2017-02-08
Update snippets.md

davidt92:
error ColormakerRegistry is ColorMakerRegistry


arose:
Hi @davidt92, I recently changed the naming, so the lowercase version is now correct, see this commit e02dc28711415026a1ba73f25da64ec8679f9446.

Issue #261, opened 2017-02-28
Orthographic zoom broken


fredludlow:
Zooming in orthographic mode seems to be broken by 0afba24. The previous zooming code detected orthographic mode by checking `camera.fov === undefined`
https://github.com/arose/ngl/blob/ad550c7abdfda7da9a37b223512d976cf8dd7b6a/src/controls/trackball-controls.js#L237

I've got a patch with a simple fix, though there may be a cleaner way, will send a PR

fredludlow:
Relatedly, and not so much a bug as a personal preference: In orthographic mode, translation moves the view a different amount depending on the zoom level. Would it be possible for translation to move 1:n (preferably 1:1!) with the mouse pointer at any zoom level in orthographic mode? (i.e. if my pointer is over a carbon and I right-click and drag the carbon should stay under the pointer).

fredludlow:
Sorry, engage brain - I guess the sums work the other way round actually: in my ideal world, screen pixels are scaled to Angstroms and the camera moved the appropriate number of pixels. I had a quick play but couldn't figure out how to go from screen units to viewer units

arose:
> if my pointer is over a carbon and I right-click and drag the carbon should stay under the pointer

that would be nice indeed

arose:
The orthographic zooming should be fine now. Could you please make a new issue for the sync of scene translation and mouse movement

PR #262, opened 2017-02-28
Bugfix for arose/ngl#261 - orthographic camera zoom broken

fredludlow:
See #261


arose:
thanks, there are more camera related things currently not working - the code is in flux...

PR #276, opened 2017-03-15
Sync mouse movement with translation

### LINKED ISSUE #275: Sync scene translation with mouse movement
Creating new issue (previously comment in #261)

When right-click dragging to translate, the scene should move the same number of pixels as the mouse pointer. For orthographic camera need to calculate pixels/Angstrom, for perspective camera maybe use halfway between near and far frustum to define pixels/Angstrom scaling.

fredludlow:
I wrote this to fix #275, then saw your commit 8de48d4. You might prefer to rewrite rather than merge, but have a PR anyway and merge/close as you see fit!


arose:
that is super neat, thanks!

PR #277, opened 2017-03-16
Build ES module file

ppillot:
Changes to the build process to produce an ES module that can be consumed by bundlers (such as rollup and webpack).


arose:
see also https://github.com/rollup/rollup/wiki/pkg.module

arose:
@ppillot just to make sure. Is this ready to go?

ppillot:
I think it is. I've tried it on my web application, using webpack and it works (no error).
I haven't noticed any tree shaking effect. I suppose it works fine, at least it works as the other es modules I use.

arose:
> I haven't noticed any tree shaking effect.

I would expect the treeshaking when creating the bundle of the application that consumes NGL.

> I suppose it works fine, at least it works as the other es modules I use.

That is good enough for me :)

ppillot:
Tree shaking is indeed sort of working: when importing only the used objects in NGL, the module's size is decreased... by 10kB (that's 2kB less gzipped)

arose:
I think the way I import and register parsers, representations and colormakers prevents treeshaking for them. Unless you edit the ngl.js entry point you always get almost everything of the library.

Issue #274, opened 2017-03-15
Bug with selections and what might be a caching issue with atomsets generated selestrings


ppillot:
I've noticed a strange bug related to successive selection operations and maybe the use of seleStrings.
Here is the test case procedure to replicate the bug:
- open page embedded.html in example folder
- paste in the console the following script
```javascript
stage.removeAllComponents()
stage.loadFile('rcsb://1GZX').then((comp) => {

comp = stage.compList[0]
    comp.addRepresentation('ball+stick',
          {
            sele: 'all'
          })
    repr0AS = comp.structure.getAtomSet() //all atoms
    console.log('Whole set size:', repr0AS.size())
    
    as = comp.structure.getAtomSet(new NGL.Selection('protein')) //protein atoms only
    console.log('Protein set size:', as.size())

    sel = comp.structure.getAtomSet(new NGL.Selection(':B')) //chain B atoms only
    
    as.difference(sel) //protein and not chain B
    console.log('Protein and not chain B set size:', as.size())
    
    repr0AS.difference(as) //chain B or not protein

    comp.reprList[0].setSelection(repr0AS.toString())
    console.log('Chain B or not protein set size:', repr0AS.size())
    
    comp.reprList[0].setSelection('protein') //should be protein but is protein and not chain B
    console.log('Protein set size:', comp.structure.getAtomSet(new NGL.Selection('protein')).size())

})
```

The output from the script is : 
```
Whole set size: 4769
Protein set size: 4384 <--
Protein and not chain B set size: 3261
Chain B or not protein set size: 1508
Protein set size: 3261 <--
```
After the script has performed, each command using the `protein` set returns a wrong set (eg: setSelection('protein'), structure.getAtomSet(new NGL.Selection('protein')) )

ppillot:
A further insight: when the string in the line `comp.reprList[0].setSelection(repr0AS.toString())` is replaced by a string such as `:B and not protein`, the correct behaviour is observed.
I've checked also with a reduced selestring such as `@0,1,2,4383` in case it is related to the previous bug with cartoon/tube representations and extrapolation. I've noticed that the expected atomSet size is returned (4), but the representation displays the whole protein...
```
stage.removeAllComponents()
stage.loadFile('rcsb://1GZX').then((comp) => {

comp = stage.compList[0]
    comp.addRepresentation('ball+stick',
          {
            sele: 'all'
          })
    repr0AS = comp.structure.getAtomSet() //all atoms
[...]

PR #279, opened 2017-03-18
Fixed Bitset.flip_all()

ppillot:
takes into account bit set length


arose:
Cool, thanks for tracking this down. Does it fix #274?

ppillot:
I don't think so... I've fixed it because it created a problem: I relied on `flip_all()` to perform an inversion of selection (such as "select not selected") and the set I get had an increase in size, preventing it to be compared with other sets.

ppillot:
Updated PR fixes #274 (would have preferred to make a separate PR, but got mixed between commits...)

arose:
Happy to merge it if you can remove `this.atomSetCache[ seleString ].clone();`.

ppillot:
Done!

arose:
thanks!

ppillot:
My pleasure!

PR #280, opened 2017-03-21
Added ease-out to animation

fredludlow:
I really like the (new?) animation stuff. I added an ease-out effect to it:

http://fredludlow.com/ngldev/examples/webapp.html

As always, take or leave as you see fit!


arose:
Nice idea! I used `smoothstep` instead, see b3acdf58881af51fdc43a2efaba7cdb22e26a45b. What do you think?

fredludlow:
I like!

PR #285, opened 2017-03-29
VolumeSurface contour display style

fredludlow:
As mentioned in #236. Added a contour option to volume surfaces that produces sparser meshes. I've also removed setAttribute method from the ContourBuffer class - I think this is redundant now (I've tested mol surfaces and volume surfaces with this removed, is there anything else that would use it?)


arose:
thank you

PR #287, opened 2017-04-01
fix distance repr setParameter labelColor 0x000000

ppillot:
Bug in distanceLabel Representation : the label color can not be updated to black using the method setParameter with 0x000000 as a color value because it is tested as false in a conditional.

There is also a linting problem in structure.js related with pickingColor not defined


arose:
thank you!

Issue #309, opened 2017-05-21
new bitarray class missing functions from fastbitset


ppillot:
I rely heavily on several functions from atomset objects based on the previous fastbitset implementation. For example, it allows me to keep track of each representation for each atom, to hide a portion of the molecule, etc... With the [replacement by the newer bitarray class](https://github.com/arose/ngl/commit/ea400a722367f9fa4e0178550db7acc8f6ecc762) my program is broken (methods as intersects, intersectionSize, union are absent).
I am about to replicate the previous functionalities, but I am wondering at what is the best: a PR to NGL or separate utils for my own needs?

arose:
Sorry about that. A PR is most welcome :) Unittests would be great!

arose:
Please see if 8250a08a54fd1e4c3da729cbc476fd1e73c02887 solves it for you, thanks!

ppillot:
Yes! I've checked and it's perfect!
Thanks @arose

PR #310, opened 2017-05-21
bitarray.js improvements

ppillot:
Here is the PR corresponding to issue #309
I have also fixed isAllclear (was not working) but isAllSet is not working as expected (I suspect _isRangeValue is failing)
(sorry I am lost in git and can't get rid of ngl.js in dist/ in that commit 👎  )


ppillot:
I added unit tests showing that _isRangeValue is buggy

arose:
thanks, used you PR as a starting point for this commit: 8250a08a54fd1e4c3da729cbc476fd1e73c02887

PR #318, opened 2017-05-31
Trilinear Interpolation

fredludlow:
At some point I'll do something that isn't surface-related...

I added some code for trilinear interpolation of volume-colored surfaces (cribbed from the Wikipedia page). The example `color/volume2` displays a ligand, hf.dx (electron density) and esp.dx (electrostatic potential). Without interpolation the colours don't change quite as smoothly.

In terms of performance, the existing example (color/volume) is not noticeably slower on my (not very powerful) PC, but very happy to add a parameter to turn it on/off if you'd like? 

Can also delete the example if it's not worth the additional bytes.


arose:
Thanks, looks good and I don't see performance problems. The bottleneck with coloring is in the chroma.js library anyway...

Issue #317, opened 2017-05-31
Can't build examples in current master


fredludlow:
Sorry you're probably in the middle of redoing this - but at the moment I can't open any examples in webapp.html using the current master. I think this is because I can't run the slimerjs gallery script (which builds `..build/gallery/scripts.json`) due to my dev server being a VM without WebGL support.

fredludlow:
Oh, I can open `examples/test.html?script=x/y` :)

arose:
yes indeed :) or `examples/webapp.html?script=x/y`

`..build/gallery/scripts.json` is just used to populate the menu in `examples/webapp.html`

we could add a non slimerjs script to make `..build/gallery/scripts.json`

arose:
list with examples script now build automatically, see a40e1ffa0fc35eb76027d0dfa310cdb462fd3b3f

PR #319, opened 2017-06-01
Build examples.json without building gallery

fredludlow:
This is a not very clean fix for #317 - have a look and see if you want to merge it or if you'd prefer something nicer! You can call `npm run gallery images:false` and it will just build the json file. 

I tried to refactor to remove slimerjs dependency but things like `fs.list` aren't in node's fs library it seems


arose:
thanks, but created a node script, see a40e1ffa0fc35eb76027d0dfa310cdb462fd3b3f

PR #322, opened 2017-06-08
bugfix for function getRepresentationsByName RegExp was not working

rfaehr:
The current getRepresentationsByName function does not work for regular expressions.

The return value of typeof RegExp is object, so the function jumps into the else clause and tries to access a member repr of name which not exists and assigns undefined to reprName. In the following every representation is added to the list. An additional instanceof condition in the if statement should do the work.


arose:
thanks, good find!

Issue #306, opened 2017-05-19
Error when attempting to change DistanceRepresentation colorScheme through setParameters


martingraham:
I've managed to find out it's a one-character thing (if I'm right)

https://github.com/arose/ngl/blob/master/src/representation/distance-representation.js#L314

`cylinderData.color = bondData.color1;`

should be

`cylinderData.color = bondData.color;`

as the bondData object doesn't have the property color1, it's color

arose:
Thanks, good find. Fixed, see 8ad97223d07340096ad9d77c57b7ec0eda9d2b5e

PR #324, opened 2017-06-13
2x 1-char fixes: color1 -> color

fredludlow:
Two tiny typos analogous to #306: `bondData.color1` -> `bondData.color`


arose:
Thank you @fredludlow!

PR #337, opened 2017-07-03
Stl writer

ppillot:
new class for creating and downloading STL files generated from a surface object

```javascript
molsurf = new NGL.MolecularSurface(structure)
surf = molsurf.getSurface({type: ‘av’, probeRadius: 1.4})
stl = new NGL.StlWriter(surf)
stl.download(‘my_file_name’)
```


arose:
thank you @ppillot

arose:
Hi @ppillot, just added a Writer base class (sharing some code and defining an interface) and updated the PdbWriter to use it. Would you mind doing the same for the StlWriter? Also, if you could add a simple test for the stl writer, that would be great! Thanks!

ppillot:
Hi @arose, yes sure. I am currently traveling and won't have a suitable working environment in the next few days, so don't expect it too soon...
What kind of test are you thinking about? (eg: stl file correctness? vertices processing?)

arose:
> What kind of test are you thinking about? (eg: stl file correctness? vertices processing?)

one option would be to also write a stl parser (like https://github.com/mrdoob/three.js/blob/master/examples/js/loaders/STLLoader.js) and use that to check number of vertices

PR #342, opened 2017-07-06
Stl writer v2

ppillot:
StlWriter refactored as a class. This PR includes a simplistic test on fake surface data (but it indeed revealed a computation bug)


arose:
thanks @ppillot!

PR #344, opened 2017-07-09
MolecularSurface getSurface() change to allow restriction to a given atomSet

ppillot:
getSurface() was wrapping the whole structure. This modification allows to provide an atomSet as parameter and to get a surface wrapping the corresponding atoms.


arose:
hmm, not too sure as you can do `MolecularSurface(structure.getView(new Selection('protein and :A')))`

ppillot:
Ok, great to know about this one!
I am currently building an interface to create surfaces from a given structure. The MolecularSurface object is instanciated with the whole structure and then individual surfaces are created as components. I guess that having a unique MolecularSurface object may be a marginal improvement compared with creating different MolecularSurface for each surface created.
Anyway, I think I am heading to a dead end with this strategy as such computed surface representations are ultimately disconnected from the structure and can not be coloured with structure dependent colorSchemes.

PR #360, opened 2017-08-03
SdfWriter Class

fredludlow:
My colleague Paul's been using the SdfWriter code this week without any obvious bugs so far, so here's a PR - comments welcome (for instance, not sure if I should have added formalCharge as an AtomProxy attribute?)


arose:
Thanks Fred!

There is an argument to be made to add the formalCharge to the Atomtype instead but that is not well developed so I find it ok to have it as part of the AtomStore.

arose:
for atom typing see https://www.cgl.ucsf.edu/chimera/docs/UsersGuide/idatm.html

PR #361, opened 2017-08-03
Reinstating parser test code

fredludlow:
I realised while writing tests for the SdfWriter that the parser tests weren't actually executing (I guess because Parser.parse changed to return a promise rather than accept a callback) this should re-enable them. 

One of the cif test fails - I've not fixed it as not sure what correct behaviour is, let me know and I can fix it
(Left out tests-sdf-writer.js as it would conflict with previous PR!)


arose:
whoops, thanks!

arose:
fixed test, see 3a11e511dd19b069fefa3e714f4b6dafc4827e01

Issue #354, opened 2017-07-21
Line representation including waters/metals (and pickable?)


fredludlow:
Apologies this is a bit of a "splurge thoughts into a text box" - I'm wondering what the best way to do this is:

I'd like to display a protein in line representation with waters and metals shown (metal-organic bonds might not always be defined - at the moment metals without bonds don't appear at all in line mode). The way pymol and others do this is to draw a cross (pymol draws 3 lines, parallel with the axes, intersecting at the point, AstexViewer draws a 2d cross in plane with the screen). 

The Point representation with useTexture and sizeAttenuation true is nearly right, but its size is specified in screen units not Angstroms so the size is fixed as you zoom in and out, which isn't quite what I'm after. 

I'd also like to be able to pick atoms from a line representation, which doesn't seem possible at the moment?

Maybe some kind of line+point representation, using a cross as the texture for isolated atoms with size specified in scene units? If the bonded atoms were also shown as 1-pixel point circular representations (in screen units) you'd be able to pick atoms too, which would be a bit of a bonus.

Alternatively, I might be massively over-complicating this and there's already a way to do this...

fredludlow:
So I played around with it a bit more, and here's a handy workaround in case anyone else wants to display a line representation but also keep atoms pickable: You can add a point representation with the following params:

    {sizeAttenusation: false,
     opacity: 0.0,
     pointSize: 15}

Then hovering/picking atoms works as it detects the (transparent) point representation. 

Doesn't solve the displaying waters/metals in line representation, but that probably needs its own new representation type.

fredludlow:
After a few weeks otherwise engaged I made a cross representation:
http://fredludlow.com/ngldev/examples/webapp.html?script=representation/cross
https://github.com/fredludlow/ngl/tree/cross-representation

This methods draws crosses a bit like pymol - aesthetically this isn't quite as nice as doing them in plane, but I think that would require a custom shader? (One that takes 4 vertices positioned at each atom and nudges them up/down/left/right in the screen plane?). 

What I'm aiming for is a "line XOR cross" representation, i.e. every atom is depicted, if possible as the end of a line/bond, if not forming a bond to another displayed atom then as a cross. Actually, what I'm really aiming for is a "pickable line/cross" representation (possibly using a transparent point representation as outlined above?). I'm not sure how best to structure this, so here's what I've come up with, feedback on structure, naming and whether to include at all very welcome!

[...]

PR #379, opened 2017-09-06
Add cross depiction to LineRepresentation

fredludlow:
References: #354. Let me know if there's anything else that needs adding or fixing.


arose:
thanks!!!

Issue #46, opened 2016-01-27
Extend picking to measurements (distance, angle, dihedral)


arose:
The picking code is sufficient but some representations/helpers are needed to display the measurements.

fredludlow:
I've just been looking at how I might go about adding distance measurements (particularly between atoms in different structures) and remembered this issue. 

I was thinking I'd just write a helper that for my interface that watched what had been picked and added some appropriate `Shape` objects - but if you've got any plans or good ideas about how to do it (either in NGL or externally) it would be good to know..

fredludlow:
Bumping this issue: I know we discussed this briefly, but what needs to be done to enable this within NGL? I've got some time to work on this if there's something useful I can do...? Or is it something that fits better at a higher level (outside of NGL)

If it were going to be part of NGL itself:
* Who owns the representations? (if they're measuring distances between atoms in different structures)
* Should the user specify "distance picking mode" then pick some atoms? Or have a concept of selected atoms (and order they were selected) so user selects first and then invokes 'addDistance' which looks at which atoms are selected.

Sorry that's a load of questions and no concrete suggestions ;)

ppillot:
I like the way it is handled in Jmol: double clicking an atom starts picking and a dashed line is displayed between this atom center and the cursor. When the cursor hovers another atom, the dashed line is magnetized and attach to this atom while a distance is displayed. The representation is definitely added when the user validates by double clicking on the target atom.
Measures are canceled by clicking outside the representation.
Angles are measured by making a single click on the "angle" atom.

arose:
A first step would be adding AngleRepresentation and DihedralRepresentation, those could be used to efficiently bring a lot of measurements on the screen, limited to be within a structure.

For more interactive, user driven measurements a separate Measurements class seems better. It would not work as a StructureRepresentation but create spheres and cylinders to show measurements the user wants to see. That way there no limitation from which structure (or volume) the position comes at which some measurement object is drawn.

fredludlow:
Makes sense: I'll start with `AngleRepresentation extends StructureRepresentation` as that should fit nicely into the existing types. 

There's a couple of places to avoid duplication: 
* Distance|Angle|DihedralRepresentation will share a lot of parameters (labelSize, offset etc) so having an intermediary `MeasurementRepresentation` type might be good?
[...]

PR #401, opened 2017-10-11
Angle repr

fredludlow:
See #46 

I've included some changes to distance representation to make label handling consistent. The way opacity (for the various different buffers) are handled doesn't seem quite right but it's a compromise between exposing absolutely everything as a separate parameter and hiding details - please feel free to completely change that (or any other bit for that matter)


arose:
thanks!!!

hainm:
can you make beta.4? thanks all.

arose:
> can you make beta.4? thanks all.

done

Issue #372, opened 2017-08-24
Make structure files available for communication with other tools


j0kaso:
In order to use e.g. MDAnalysis in MDsrv, the structure file is requested. Please make it callable / available on request (like it is possible for remote trajectories).

arose:
feature request for the mdrsv datasource (https://github.com/arose/ngl/blob/master/src/datasource/mdsrv-datasource.js)

arose:
@j0kaso can you please describe again what we need here, on the server as well as in NGL (in the mdsrv datasource). Thanks!

j0kaso:
MDsrv needs the trajectory and the structure file when MDAnalysis should be used. So far, the trajectory file loaded through the data source is callable by the MDsrv (via the GET/POST request as here: https://github.com/arose/ngl/blob/master/src/trajectory/remote-trajectory.js), the "location"/path of the structure file is not. Both are needed (and could come from a different location). I assume the path of the structure file would be needed to be saved (maybe difficult for drag'n'droped structures?). Afterwards, MDsrv needs to GET this location and forward it to MDAnalysis. Hope this helps.

arose:
Can you suggest an url you can consume in mdsrv that includes the required info?

j0kaso:
similar to traj: "/traj/numframes/<root>/<path:filename>" --> "/struc/<root>/<path:filename>"
I assume it should be fine with this (could also think that further information as number of atoms, ... could also be of interest in the future and with regards to further analysis)

arose:
doesn't the structure path have to be send with every traj request?

j0kaso:
right, otherwise it would be mixed up with other trajs... so just the info after the traj?: 
`/traj/frame/<int:frame>/<root>/<path:filename>/<root2>/<path:filename2>`

arose:
or as a GET param?

j0kaso:
the traj information is send via GET so it would be easier to continue with this. still together, otherwise it would be mixed up with other structures (thanks for the reminder)

arose:
changed the mdsrv datasource such that any query string from the original file path gets appended to requests for frames

```
original path: file://cwd/data/md.xtc?foo=bar
frame path: traj/frame/42/cwd/data/md.xtc?foo=bar
```

you can put anything in the query string

can you work with that @j0kaso ?

arose:
see f0c7d126ab0456f5e352b11c5fead1dcd1a6ca68

j0kaso:
works - thanks for including it!

PR #402, opened 2017-10-12
structure path availability

j0kaso:
structure file path available with trajectory file for other tools (mdsrv) via GET request
(solving issue #372 )


arose:
thanks!

PR #403, opened 2017-10-16
Remove bash-ism from package.json, (fixes build error on Ubuntu)

fredludlow:
Tiny PR - for some reason the expansion {*.vert,*.frag} in the "ts" script fails on both my ubuntu boxes (though it works fine on CentOS)


arose:
thanks

PR #405, opened 2017-10-25
MMTF formal charge as signed int

fredludlow:
Replaces unit8 with int8 for storing formalcharge in mmtf parser


arose:
Ouch, how did that happen... Thanks

PR #408, opened 2017-11-02
Added astexviewer mouse preset

fredludlow:
Because it's what my users are used to. Hope that's okay!


arose:
sure, that's way it is configurable :)

PR #411, opened 2017-11-09
Fix missing radii / odd surfaces

fredludlow:
There's a small typo molecular-surface.js which results in all atoms having radius 1.0A for surface generation - changed it to use vdw. I haven't looked how this would work with custom radii.


arose:
Whoops, that was due to my recent changes. Thanks for noticing!

PR #413, opened 2017-11-17
Small measurement fixes

fredludlow:
This fixes two bugs in the angle/dihedral representations

- Angle representation with an empty `structureView` threw error (`setVisibility` was trying to set attributes of buffers that had not been created)
- Missing lines in dihedral representation
- Incorrect final coordinates for dihedral dashed lines

It also adds a `getFixedLengthWrappedDashData` which wraps the dashes along a line/cylinder path and uses it in dihedral/angle represnetations . I've put this in as a new function as you wouldn't want this for interaction repr or the distance repr. It could be improved a little by only wrapping when one line section starts at the end of the previous one (and various other nice little touches like stretching over total length of continuous sections) but as it is it makes the angle and dihedral representations look a bit better I think.

### DIFF
--- a/src/geometry/dash.ts
+++ b/src/geometry/dash.ts
@@ -145,3 +145,106 @@ export function getFixedLengthDashData<T extends CylinderBufferData|WideLineBuff
 
   return d as T
 }
+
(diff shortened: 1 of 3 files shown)


arose:
thanks, agree, makes them look better!

PR #428, opened 2017-11-30
More contact angle updates

fredludlow:
Bit more tweaking interaction thresholds and angle calculations:

* Checking donor and acceptor angles (D'-D-A and D-A-A') all possible heavy atoms D' and A' are considered. This particularly helps avoid some of the unrealistic angles seen from tetrahedral donors. Same change implemented for halogen bond acceptors (though I'm not sure how often you see a furan- or ether- Br interaction!)

* Acceptor angles: 
  * Checking angle acceptor now only checks for angles less than (ideal - tolerance). This allows us to set a tighter threshold for acceptor angle (to avoid very small A'-A-D angles) but also keep  C=O-HN where the angle around the oxygen is about 180.
  * I've effectively turned off the dihedral check for sp2 acceptors (setting default tolerance to 90) to be consistent with the above.

* Donor dihedral (sp2 donors): I've reduced the tolerance to 30 (this is actually a bit of a misnomer as it's an out-of-plane measurement rather than the dihedral form the Rose-Dill paper

Let's hope I've not messed anything up this time!


arose:
thanks, renamed dihedral to planeAngle in contact params (see 85b1944163fb5fe636ba72d0dfd2df5e25e69b7d)

PR #434, opened 2017-12-05
Interactive crosslink example 1.1

martingraham:
Here you go (I think)

- redoing pull request as I'd added in a package-lock.json file to my repository by mistake

Caveats:

I introduce a library dependency - d3.js - it's my preferred library for dom/array manipulation - I use it for adding the slider, working out the data range for the link scores etc. It could be replaced by pure javascript I suppose, but then again so could most things :-)

The slider controls don't disappear when loading another example (but that seems to happen with other examples like the ligand-viewer). Maybe need a line in the load script code to get rid of any sibling dom nodes to the viewport?


arose:
Thanks!

I'd rather not introduce an extra library. Would it be much work to change it?

PR #436, opened 2017-12-06
Cross-linking interactive example 1.2

martingraham:
Okay, another attempt with no d3 this time :-)

I had to add eslint and a couple of others to the package.json as it wouldn't build without them (I think the standard lint package has dependencies on them)


arose:
thanks!

PR #441, opened 2017-12-13
Carboxylate detection and sulfonamide protonation

fredludlow:
Fixes the case of acetate anion (not detected as carboxylate) and incorrectly assigning positive charge to sulfonamide nitrogen


fredludlow:
Bundling two other changes:
Fix in BondOrderTable for ASP carboxylates
Add hover tooltip for astexviewer mouse preset

arose:
> Fix in BondOrderTable for ASP carboxylates

good find, thanks

PR #450, opened 2018-01-18
Add 'mol' as supported file extension for SdfParser

fredludlow:
As the title suggests - allows opening of .mol files directly from UI or via API without specifying 'sdf' type


arose:
thanks

PR #451, opened 2018-01-18
Fix for fixedSize labels in orthographic view

fredludlow:
The `fixedSize` labels are great, but they behave strangely in orthographic projection. (The perspective adjustment is still applied so the ones close to the camera are tiny and the ones far away are huge). This seems to fix it, though, as always, there may be a better way!


arose:
good find, thanks

Issue #461, opened 2018-02-02
Annotations outside the stage


gbayarri:
Hi,

I'm trying to add annotations to the proteins of a PDB file following the NGL gallery example named _interactive/annotation_.

The problem is that the annotations are shown outside the stage and in an absolute position respect to the browser. They interact with the viewport (i.e. if we rotate the object, they rotate, if we zoom, they zoom) and the positions seem to be correct between them, but they have an offset. Here you have a screenshot:

![screenshot](https://user-images.githubusercontent.com/19928100/35739216-38d32cc2-0831-11e8-8aab-0041c19e9310.png)

And here an interactive example:

http://mmb.irbbarcelona.org/webdev/test/annotations.html

The code is practically the same of the gallery example.

Tested with ngl.1.0.0-beta.7 and ngl.0.10.5-21.

Anybody knows how to fix that?

Thanks!

ppillot:
It seems related to css positioning. Theses annotations have a fixed position setting and depends on the bottom coordinates for their vertical positioning. By shrinking vertically your window, you can move them back in place (just for testing).
I manually fixed it by setting the viewport container and each annotation block, `position` as `relative`.
I've checked the code source, and the `fixed` position setting is hardcoded there.
 I couldn't find back where the example is in the gallery. Can you give the link so that I could check what the css are ?
I guess this problem doesn't arise as long as the bottom part of your viewport is at the bottom of your web page.

arose:
@ppillot the example is here http://nglviewer.org/ts2/?script=interactive/annotation

@gbayarri does it work with ts2 branch?

gbayarri:
@arose It doesn't work. Actually, I shrunk the browser in the example you provided and the annotations are out of the viewport:

![screenshot](https://user-images.githubusercontent.com/19928100/35797956-a17712ac-0a61-11e8-9daf-4c7f0d0a3e16.png)

@ppillot I'm sorry, but I'm not able to find the css in order to change them. I've checked them via browser inspector and changing the position to relative, the annotations appear at the correct position, but they are still outside the viewport when zooming.

ppillot:
The behaviour you are referring to in that test may have a different cause as it happens after resizing. Positions are correct but the boundaries to make the annotations visible (or not) are not updated after resizing of the viewport. Am I correct?

ppillot:
I have made a codepen to replicate the problem with a centered viewport that does not extend to the bottom of the page: https://codepen.io/ppillot/pen/EQyVwQ 
[...]

PR #466, opened 2018-02-06
 fix: relative css position for preventing bug in annotations #461

ppillot:
For this fix to work I had to add a wrapper div to canvas and annotations. Without this wrapper, we'd have to directly change the css positioning for the container provided by the user and this could lead to side effects for the user (e.g. if the user sets another positioning style for its container).
This fix doesn't solve the issue of annotations displayed outside of their container element.

### DIFF
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -572,9 +572,9 @@ class Stage {
    * @return {undefined}
    */
   setSize (width: string, height: string) {
-    const container = this.viewer.container
+    const container = this.viewer.container.parentElement
 
-    if (container !== document.body) {
+    if (container !== document.body && container !== null) {
       if (width !== undefined) container.style.width = width
       if (height !== undefined) container.style.height = height
       this.handleResize()
@@ -596,7 +596,7 @@ class Stage {
     }
 
     const self = this
-    element = element || this.viewer.container
+    element = element || this.viewer.container.parentElement
     this.lastFullscreenElement = element
 
     //
--- a/src/viewer/viewer.ts
+++ b/src/viewer/viewer.ts
@@ -235,27 +235,29 @@ export default class Viewer {
       rendered: new Signal()
     }
 
+    const wrapper = document.createElement('div')  
     if (typeof idOrElement === 'string') {
       const elm = document.getElementById(idOrElement)
       if (elm === null) {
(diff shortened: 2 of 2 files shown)


ppillot:
Added a new commit to the PR with a fix for the annotations outside of the viewport.

ppillot:
last fix for a small annotation overflow at the bottom of the viewport. The overflow styling is now on the user provided container (not the wrapper), as the wrapper has a few extra pixels in height.

arose:
thanks, will test tomorrow

arose:
Hi, reverted it. It does not work when the users does not supply an Element or an element id.

ppillot:
Sorry about that, I'll investigate further and make the relevant test cases.
By the way, to improve code clarity, I could create an extra object in stage: there is already `this.container`, but I would like to add `this.wrapper` as an internal reference to the additional div created for positioning the annotations.

arose:
> there is already this.container, but I would like to add this.wrapper as an internal reference to the additional div created for positioning the annotations.

sounds good, I like being very explicit, especially if it does not matter performance wise

ppillot:
Hi, what is the expected behaviour if no HTML Element or element id is provided?
I've checked with latest version and nothing is displayed on screen (but the stage object is created and includes an actual reference to a container element).

Additional question: what shall be done regarding tooltips? They are displayed on top of everything else and can overflow the viewer. I think this behaviour is nice as a cropped tooltip would not be informative, so I would favour keeping tooltips as they are.

arose:
> Hi, what is the expected behaviour if no HTML Element or element id is provided?
I've checked with latest version and nothing is displayed on screen (but the stage object is created and includes an actual reference to a container element).

`stage.viewer.container` should be a plain div whose style can be changed by the user

> Additional question: what shall be done regarding tooltips? They are displayed on top of everything else and can overflow the viewer. I think this behaviour is nice as a cropped tooltip would not be informative, so I would favour keeping tooltips as they are.

totally agree, no cropped tooltip

ppillot:
> `stage.viewer.container` should be a plain div whose style can be changed by the user
Meaning this div should be appended to the DOM at some stage? At present there is no code for handling that in viewer.ts (a div is created but not appended to the body of the page)

Issue #391, opened 2017-09-27
Custom labels


fredludlow:
Not urgent, but while I think of it - there might be some use for a mechanism for constructing custom label text e.g.:

e.g. handlebars.js / jinja2 syntax:
`{{ chain }}:{{ resno }} {{ bfactor }}`

Or more compact/cryptic printf style, with precision
`%c:%r %.1b`

arose:
would be nice indeed! there is already sprintf (https://github.com/arose/ngl/blob/master/lib/sprintf.es6.js) in ngl which includes named arguments and can be used for number formating :)

arose:
I suppose once a new sprintf version is released (https://github.com/alexei/sprintf.js/pull/161), we can have this goodness https://github.com/fredludlow/ngl/tree/sprintf-labels, right?

fredludlow:
Yup, that's the idea, or we can just reference a specific git commit in the sprintf.js repo?

arose:
> or we can just reference a specific git commit in the sprintf.js repo?

fine by me too

PR #467, opened 2018-02-07
Sprintf labels

fredludlow:
See #391. Updated the representation/label example to use a custom format.


arose:
very nice, thanks!

PR #470, opened 2018-02-08
fix: relative css position for annotations placement #461

ppillot:


arose:
Thanks, this looks good to me, tried to break it, but couldn't :)

I added a `white-space:nowrap` to the annotations so there is no wrapping on the right side of the container.

@gbayarri does it work for you now?

gbayarri:
It works perfect!

Thanks.

Issue #446, opened 2017-12-24
surface mesh lineWidth


giagitom:
Hi, I am actually using version 1.0.0-beta.7. It seems that the surface mash line width keeps a fixed value if I try to change it using setParameters({lineWidth:1.5})

fredludlow:
I've been meaning to take a look at this as I'd quite like it to work too! The short answer is that linewidth is ignored when you draw simple lines on Windows (possibly other platforms too). The `line` representation gets around this by drawing rectangles (using `WideLineBuffer`). 

For surfaces, there are two ways to get something that looks like a mesh

1) `wireframe` mode: uses three.js's `LineSegments` for rendering, I don't know enough about this bit of the code to offer any useful suggestions about replacing this :)
2) `contour` mode: draws lines, but there's no reason why it shouldn't draw rectangle (@arose  hand't implemented the WideLineBuffer when I was working on that bit). 

Alex, to get linewidths in contour mode I can replace ContourBuffer with WideLineBuffer and map the {position, color, index} attributes to {position1, position2, color1, color2} attributes that WideLineBuffer expects. I'm not quite sure where the best place to do this is:

1) Create a general function for converting mesh-style data {position, index, color} -> the format for WideLineBuffer {position1/2, color1/2} 
2) Generate correct format directly when building the contours in marching-cubes (avoids the extra intermediate objects but can't easily reuse)

Or maybe there's a better way that gets both `contour` and `wireframe` rendered as wide lines?

arose:
I had not thought of the contour mode. It would be great to have that @fredludlow! I think fully switching the ContourBuffer to WideLineBuffer is best, including generating the correct format directly when building the contours in marching-cubes.

giagitom:
Hi, any news on this?

fredludlow:
There's an open PR, but actually playing with it on big maps the performance suffers somewhat (each line segment becomes two triangles, so the memory usage expands accordingly). I'm not sure what the best thing to do with it is tbh.

giagitom:
Ok thanks @fredludlow

PR #472, opened 2018-02-14
Use WideLineBuffer for contour surfaces

fredludlow:
Addresses #446, but probably want to take a good look at results before deciding whether to merge

I've replaced the `ContourBuffer` with `WideLineBuffer` in `MolecularSurfaceRepresentation` and `SurfaceRepresentation`. 

This version does it with a simple function for converting mesh data to line data (in `surface-utils.js`) rather than the way we discussed in #446 . It would be a little more efficient to write the correct output in marching-cubes, but it seemed to need a lot of other changes to the Surface class and the two representation classes to accommodate doing it that way. Performance seems to be okay doing it this way... What do you think?

The representation/molecularsurface-av and test/xray demos are a good comparison of changed behaviour:
Xray:
http://nglviewer.org/ts2/?script=test/xray
http://fredludlow.com/ngl/contour-width/examples/webapp.html?script=test/xray

Surface:
http://nglviewer.org/ts2/?script=representation/molecularsurface-av
http://fredludlow.com/ngl/contour-width/examples/webapp.html?script=representation/molecularsurface-av
NB changed the ligand surface to linewidth:2 to demo feature

At linewidth: 1.0 it looks a little jagged, maybe default should be higher? Or is there a way to render narrow lines a bit more smoothly?

### DIFF
--- a/examples/scripts/representation/molecularsurface-av.js
+++ b/examples/scripts/representation/molecularsurface-av.js
@@ -30,5 +30,4 @@ stage.loadFile('data://2vts.mmtf').then(function (o) {
     -23.85, -24.05, 7.95, 0,
     -27.16, -8.65, -63.38, 1
   ])
-  stage.setParameters({sampleLevel: 2})
 })
--- a/examples/scripts/representation/molecularsurface-av.js
+++ b/examples/scripts/representation/molecularsurface-av.js
@@ -11,6 +11,7 @@ stage.loadFile('data://2vts.mmtf').then(function (o) {
     contour: true,
     colorScheme: 'element',
     colorValue: '#0f0',
+    linewidth: 2.0,
     useWorker: false
   })
   o.addRepresentation('surface', {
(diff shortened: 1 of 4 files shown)


arose:
Thanks for taking a stab at this @fredludlow!

http://fredludlow.com/ngl/contour-width/examples/webapp.html?script=test/xray seems broken.

fredludlow:
Hmm, it's working for me - is the example broken or server inaccessible?

fredludlow:
@arose - is the example still broken for you? (Still working here...)

arose:
it works, i wanted to add reuse of gl buffers for wideline shader because it is somewhat slow when recalculating a mesh when moving around

fredludlow:
I'm closing this as I think it's not very performant and I've stopped using it locally, but can resurrect it if there's interest later. There's also a bug with not reallocating arrays when the input data gets bigger (so maps get truncated).

PR #474, opened 2018-02-20
avoid AtomMap to be overwritten when atom names in the input file have same atom names

sunhwan:
[2gp_m1.mol2.zip](https://github.com/arose/ngl/files/1741313/2gp_m1.mol2.zip)


arose:
Thanks @sunhwan. Could you explain why this is exactly a problem? I fear that your code will not scale nicely to very large structures.

sunhwan:
I added an example which motivated this fix. When the atom names are identical in the input structure file, the AtomMap's dict key gets overwritten. Given AtomMap is the only way to access certain atom information (name, element, etc), I think we need to rename atoms in such case.

Thinking about this again, I realized protein residues have same atom names, so this is not a good approach. 

I figured I should query atomStore to get atomTypeId first and use that to query AtomMap.list. I will close this pull request.

PR #481, opened 2018-02-26
Kin code ribbon & balls addition

vbchen:
VBC's initial attempt at adding ribbonlist and balllist processing to the kin parser.
Also added parsing of a few more kin keywords.
I changed a few variable names to aid in my learning how the code works.


arose:
As far as I can tell you have done three commits, right? The rest of the commits are just from you updating your branch with master, right?

This looks very messy. How about I add what I had done for the kin parser to the ts2 branch (which is where development currently happens) and after that you would redo you changes on top of ts2 and do a new PR.

vbchen:
Yeah, I did the last three commits.  

Sure, we can use the ts2 branch, whatever you think is best, thanks!

arose:
Merged the 'kin' branch into 'ts2' which I merged into 'master'... So please do your changes on top of the master branch. Thanks!

See the kin parser example here http://nglviewer.org/ngldev/?script=parser/kin

PR #490, opened 2018-03-06
fix: concatstructure forgets element

csulok:
Noticed an issue with concatstructure dropping an atom's element as parsed from the input file, and I traced it to this difference between e.g the SDFParser. Could you look at the patch and let me know what else would be needed to merge the change?

Could be tested with the fluorine of https://www.rcsb.org/ligand/5SZ
Here's a codepen showing the fluorine is white as a result: https://codepen.io/anon/pen/bLPXrx?editors=0010


arose:
good find, thanks!!!

PR #492, opened 2018-03-07
fix: concatstructure forgets altloc, occupancy, bfactor

csulok:
new pull request after https://github.com/arose/ngl/pull/491


arose:
thanks again

PR #501, opened 2018-03-20
updated kin.js example

vbchen:
Updated the example kinemage.
Updated the kin.js code with the additions to kin-parser.


examples/scripts/parser/kin.js
```diff
@@ -1,42 +1,62 @@
...
-      if (!dotList.master.includes(master)) return
-      for (var i = 0, il = dotList.position.length / 3; i < il; ++i) {
-        var i3 = i * 3
-        var x = dotList.position[ i3 ]
-        var y = dotList.position[ i3 + 1 ]
-        var z = dotList.position[ i3 + 2 ]
-        var r = dotList.color[ i3 ]
-        var g = dotList.color[ i3 + 1 ]
-        var b = dotList.color[ i3 + 2 ]
-        shape.addPoint([ x, y, z ], [ r, g, b ], dotList.label[ i ])
```
arose: You should actually use the shape approach here to get support for picking/hover labels

examples/scripts/parser/kin.js
```diff
@@ -1,42 +1,62 @@
...
+    })
+
+    var positionArray = []
+    var colorArray = []
+    kinemage.ribbonLists.forEach(function (ribbonList) {
+      if (!ribbonList.masterArray.includes(master)) return
+      positionArray.push(...ribbonList.positionArray)
+      colorArray.push(...ribbonList.colorArray)
     })
+    if (positionArray.length > 0) {
```
arose: For now it is ok to use a buffer here but we should add a method `shape.addTriangle` to the `Shape` class to get support for picking/hover labels here as well.

arose: Here is the commit (https://github.com/arose/ngl/commit/545cc9387777205acdb7cf71b8e3f430816ad421) where I added `shape.addPoint` and `shape.addWideline` if you want to give it a try.

examples/scripts/parser/kin.js
vbchen: Ah, ok, cool!  I'll give it a try.  That's my fault for not looking more closely at your recent commits.

examples/scripts/parser/kin.js
vbchen: Sounds good, I'll give it a go.

arose: great, thanks!

PR #502, opened 2018-03-21
FIX: Measurements defined with atom indices respect component selections

fredludlow:
You can specify measurement parameters (`atomPair`, `atomTriple`, `atomQuad`) with either integer atom indices or string selection expressions. Selection expressions were/are evaluated in context of component's structureView, so automatically respect selection on component. 

This change ensures consistent behaviour when using integers to specify atoms for measurement representations.


arose:
great, thank you!

Issue #458, opened 2018-01-30
A question about angle and dihedral repr


giagitom:
Hi, is it currently possible to draw dihedral and angle repr in a pymol like fashon? Like this I mean:

![image](https://user-images.githubusercontent.com/20627977/35581400-f505e4f4-05eb-11e8-987e-a3bc6f532e20.png)

arose:
the angle yes (http://nglviewer.org/ts2/?script=representation/angle), but the dihedral looks a bit differently (http://nglviewer.org/ts2/?script=representation/dihedral)

fredludlow:
The dihedral with `{planeVisible: false, sectorVisible: false}` won't look quite the same but is closer. (Also just noticed that changing planeVisible doesn't trigger a rebuild, I'll fix that)

giagitom:
Thanks. Is it something hard to implement a parameter to hide the connectors and a parameter to draw the two lines making the sector perimeter?

PR #503, opened 2018-03-21
Dihedral fix

fredludlow:
This fixes the small bug in #458 changes to `planeVisible` didn't rebuild. Also adds a `extendLine` parameter. This defaults to true (displays as before), but if this is false the representation looks approximately like that requested by @giagitom in #458

(Apologies for the delay in doing this...)


giagitom:
Thank you very much @fredludlow !

arose:
Very nice, thank you @fredludlow

fredludlow:
Oops, I'll have a look (the logic in the point generation is getting a bit convoluted, so maybe best to refactor it anyway)

arose:
thanks!

PR #510, opened 2018-04-04
Another dihedral fix

fredludlow:
Fix for bug introduced in #503 (extendLines mode was still allocating arrays for the plane mesh)


arose:
thanks

Issue #517, opened 2018-04-10
Contacts between 2 entities only


LivC182:
Is there anyway to calculate contacts just between polymer and ligand, without showing polymer to polymer interactions or ligand to ligand interactions? 
I tried: sele:"polymer and ligand" and it dosent work.
sele: "polymer or ligand" works but I get all the interactions within the protein which hinders the visibility for the important contacts.

arose:
There is a 'filterSele' parameter in the contact representation you can use, try setting it to 'ligand'

LivC182:
Works pretty well: filterSele: ligand, gives the interactions between the ligand and all other entities(polymer, ligand, water, ion) so a combination between filterSele and sele should give only the interactions between ligand and polymer


One more thing: I am trying to change the color coding for contacts but I couldn't find any contact-colormaker.ts or a contacts.ts to see where they you init them. Is there color a parameter for each contact ?  

Thank you for your help

arose:
not yet, tracked in #435

ppillot:
I have a concern with the `filterSele` parameter in one case: it works very well for a simple ligand, but I wanted to apply it to antibodies and antigenes. if I use the "antigene" as `filterSele`, then I also get contacts that are internal to the antigene and not only contacts relevant to the epitope binding.
I was wondering if a second filter could be a good option for this scenario?

LivC182:
I had know idea 'antigene' was a recognized value for filterSele. Do you know where I can find all the values this param can take ?

ppillot:
antigene is not a proper selector, that was a (poor) way of describing the invocation of a selector. The list of all the selector accepted by NGL is in the manual at http://nglviewer.org/ngl/api/manual/usage/selection-language.html

csulok:
`filterSele` keeps contacts where at least 1 end is in the selection and I think @LivC182 would need the extra check for both ends.

See here: https://github.com/arose/ngl/blob/master/src/chemistry/interactions/contact.ts#L271

PR #519, opened 2018-04-15
Contacts: filterSele can be an array to filter contacts between 2 sets

ppillot:
This PR adresses a limitation with filterSele described in #517 
E.g.: when studying the contacts between an antigen and an antibody, you can set filterSele to the antigene. But you'll get contacts between the antibody and the antigene, as well as undesired contacts that are internal to the antibody.
In this proposal, filterSele can also be set as an Array. If that is the case, then the script will filter contacts keeping only those that are set between both selections.

### DIFF
--- a/src/buffer/doublesided-buffer.ts
+++ b/src/buffer/doublesided-buffer.ts
@@ -124,8 +124,8 @@ class DoubleSidedBuffer {
       front = this.frontBuffer.getMesh()
     }
 
-    this.frontMeshes.push(front)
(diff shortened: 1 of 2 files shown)


arose:
great, thanks!

LivC182:
Super nice. I was trying for a while to work around but I could not manage (I am a beginner in js). Just to make sure I am using it properly:

filterSele:['ligand','polymer'] only gives the interactions between ligand and polymer

PR #569, opened 2018-08-16
Look at neighbouring pixels when picking

fredludlow:
This is something I've been playing with here, and found it quite useful for easier picking of single-pixel line representations.  Basically it looks in a 5*5 rectangle of pixels centred on the pick event (starting at the middle and working outwards). 

If you compare picking in the line+cross representation example between this and the standard app you can see what I mean:
http://fredludlow.com/ngl/pick-wider/examples/webapp.html?script=representation/line-and-cross

If you think it would be better controlled by a parameter I can add that (stage.pickRadius?).

Fred

### DIFF
--- a/src/viewer/viewer.ts
+++ b/src/viewer/viewer.ts
@@ -43,6 +43,11 @@ import Buffer from '../buffer/buffer'
 const pixelBufferFloat = new Float32Array(4 * 25)
 const pixelBufferUint = new Uint8Array(4 * 25)
 
+// When picking, we read a 25 pixel (5x5) array (readRenderTargetPixels)
+// We read the pixels in the order below to find what was picked.
+// This starts at the center and tries successively further points.
+// (Many points will be at equal distance to the center, their order
+// is arbitrary).
 const pixelOrder = [12,7,13,17,11,6,8,18,16,2,14,22,10,1,3,9,19,23,21,15,5,0,4,24,20]
 
 
--- a/src/viewer/viewer.ts
+++ b/src/viewer/viewer.ts
@@ -40,8 +40,11 @@ import { testTextureSupport } from './gl-utils'
 
 import Buffer from '../buffer/buffer'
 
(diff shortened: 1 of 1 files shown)


src/viewer/viewer.ts
```diff
@@ -931,33 +934,42 @@ export default class Viewer {
     x *= window.devicePixelRatio
     y *= window.devicePixelRatio
 
-    let pid, instance, picker
+    x = Math.max(x - 2, 0)
```
arose: I think the 2 should be multiplied by `window.devicePixelRatio`

src/viewer/viewer.ts
```diff
@@ -40,8 +40,11 @@ import { testTextureSupport } from './gl-utils'
 
 import Buffer from '../buffer/buffer'
 
-const pixelBufferFloat = new Float32Array(4)
-const pixelBufferUint = new Uint8Array(4)
+const pixelBufferFloat = new Float32Array(4 * 25)
+const pixelBufferUint = new Uint8Array(4 * 25)
+
+const pixelOrder = [12,7,13,17,11,6,8,18,16,2,14,22,10,1,3,9,19,23,21,15,5,0,4,24,20]
```
arose: Could you add a comment what the order mean geometrically? Thanks.

src/viewer/viewer.ts
fredludlow: Would that change the size of the buffer that we need to read to too?

src/viewer/viewer.ts
fredludlow: Will do (It starts at the centre and looks at pixels by distance to centre)

src/viewer/viewer.ts
arose: Right, it would need to be changed. Let's forget about it.

src/viewer/viewer.ts
fredludlow: Done - are we good to merge?

PR #574, opened 2018-09-02
Generate d.ts declaration files

ppillot:
This PR contains modifications to generate correct d.ts files that could be shipped with ngl from npm.

To solve the bug with modules exporting a name defined in another module but not imported, instead of the dummy export proposed previously, a cleaner hack was used in https://github.com/arose/ngl/commit/5cfc67c99ebe153139ef0903ac0c283cf348a0ac : a comment (`// @ts-ignore`) has been added before each of these import to discard tsc checks for unused imports. A discussion of this problem can be seen at Microsoft/TypeScript#5711

The d.ts declaration generation has been added as a task in package.json: `npm run dts` produces declarations for each module in the NGL library (it targets a new "declarations" folder). It has not been added to the publish workflow.
To ship these declarations as part of the ngl package on npm, a `types` property pointing to the "declarations" folder is added to package.json and @types declarations for signal, three and chroma were moved to dependencies as per suggested in [typescript documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

If these modifications are accepted, the npm package generation should be modified to include the "declarations" folder next to the already shipped dist folder.

Finally, with this design I can import typings from NGL in my projects written in typescript with a syntax like `import AtomProxy from 'ngl/declarations/proxy/atom-proxy'`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 build/
 node_modules/
+declarations/
 
 npm-debug.log
 
--- a/package.json
+++ b/package.json
@@ -4,12 +4,14 @@
   "description": "Scalable molecular graphics for the web",
   "main": "dist/ngl.js",
   "module": "dist/ngl.esm.js",
+  "types": "declarations/ngl.d.ts",
   "scripts": {
     "lint": "npm run lint-src && npm run lint-test && npm run lint-script",
     "lint-src": "standard \"src/**/*.js\"",
     "lint-test": "standard --env mocha \"test/**/*.js\"",
     "lint-script": "standard --global stage --global NGL \"examples/scripts/**/*.js\"",
     "ts": "tsc && cpx \"src/**/*.{vert,frag,glsl}\" build/js/src/ && cpx \"lib/*.js\" build/js/lib/ && cpx \"package.json\" build/js/ && cpx \"src/polyfills.js\" build/js/src/",
+    "dts": "tsc -d --declarationDir \"declarations\" --emitDeclarationOnly --skipLibCheck",
     "prebuild": "npm run lint",
     "build": "npm run ts && rollup -c",
     "postbuild": "node ./scripts/makeScriptsList.js",
@@ -58,13 +60,10 @@
     "science"
   ],
   "devDependencies": {
-    "@types/chroma-js": "^1.3.5",
     "@types/jest": "^23.3.1",
     "@types/node": "^10.5.7",
     "@types/promise-polyfill": "^6.0.0",
-    "@types/signals": "1.0.1",
     "@types/sprintf-js": "^1.1.0",
-    "@types/three": "^0.92.15",
     "babel-plugin-array-includes": "^2.0.3",
     "babel-plugin-external-helpers": "^6.22.0",
     "babel-preset-es2015": "^6.24.1",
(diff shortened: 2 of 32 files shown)


arose:
Looks very nice! Thank you!

Is it necessary to have a separate 'declarations' folder or can the d.ts files be put in the dist folder directly? Asking because `import AtomProxy from 'ngl/declarations/proxy/atom-proxy'` looks odd with the declarations folder in the path. It does only import the type, right?

ppillot:
You're right, it's possible to point to the dist folder. In that case, the import path would look like 'ngl/dist/proxy/atom-proxy/' and all the d.ts files will also be in the dist folder. I've pointed to a different folder to separate concerns, but as I gathered from different examples on Github, there really is no rule for that:
- Vue library ships its declarations in a dedicated types folder https://github.com/vuejs/vue
- Litemol has d.ts files alongside its exported modules in dist https://github.com/dsehnal/LiteMol/tree/master/dist/js

To answer your last question, yes it only imports the type.

sgratzl:
what is the status of this PR? it would be nice to see that the typings are distributed such that other Typescript applications can use this library smoothly.

garyo:
I also have a private fork with @ppillot 's changes. Would be good to get this merged in. (I have some other pending PRs I can't submit without this one, like updating to latest three.js.)

arose:
Hi @ppillot and @fredludlow what is your suggestion?

fredludlow:
I haven't used NGL in a TS project yet, but we should get this merged. 
@ppillot, @garyo and @sgratzl - if you're all using these changes as they are in this PR then I think we should just merge.

ppillot:
Hi, I use this but I am not 100% sure it is doing the right thing because I need to import the typings in a kind of manual way (with the path to the imported typings) in my ts scripts.
Hence, there is room for improvements but it's certainly more beneficial to merge it as it is rather than wait for an eventual improvement.
In short, let's merge and let's improve it

garyo:
I can confirm this is working fine for me in a typescript app. Specifically I merged a5c7a08a from ppillot/dts3, and after running `npm run dts` and `npm run build-min` I can import using `import * as NGLViewer from 'ngl'` just fine. (I symlink the ngl dir into my project's node_modules.)

sgratzl:
> I haven't used NGL in a TS project yet, but we should get this merged.
> @ppillot, @garyo and @sgratzl - if you're all using these changes as they are in this PR then I think we should be okay to merge?

yes would be nice to merge this PR and then add missing TypeScript exports or mark some exports as `@internal` and filter them out with `--stripInternal`.

fredludlow:
Right, I'll go ahead and merge.

@arose - would it be sensible to make @ppillot  and @garyo maintainers? (in a case like this it's only you or I who have access to approve but you're busy with mol* and I don't have a lot of experience with TS)

arose:
> would it be sensible to make @ppillot and @garyo maintainers? (in a case like this it's only you or I who have access to approve but you're busy with mol* and I don't have a lot of experience with TS)

@ppillot and @garyo are you interested?

arose:
> Right, I'll go ahead and merge.

thank you @fredludlow

ppillot:
Thanks for the offer. I'd be glad to help.

garyo:
I'd be happy to help as well.

arose:
Wonderful, welcome as collaborators then @garyo and @ppillot, you should be getting an email from github.

PR #581, opened 2018-09-27
Add toJSON to buffer.ts to avoid circular refs when exporting

garyo:
The groups' children refer back to the mesh, and the picking object
has a circular structure.data attribute.

I don't believe the groups or picking are required for
serialization (e.g. model export) so this change shouldn't impact
normal use.

### DIFF
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -67,11 +67,17 @@ declare global {
     msFullscreenEnabled: boolean
     msFullscreenElement: Element
     msExitFullscreen(): void
+
+    webkitFullscreenEnabled(): boolean
+    fullscreenElement: Element
+    webkitFullscreenElement: Element
+    webkitExitFullscreen(): void
   }
 
   interface Element {
     mozRequestFullScreen(): void
     msRequestFullscreen(): void
+    webkitRequestFullscreen(): void
   }
 }
 
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -108,6 +108,8 @@ declare global {
  * @property {Color} ambientColor - ambient light color
  * @property {Float} ambientIntensity - ambient light intensity
  * @property {Integer} hoverTimeout - timeout for hovering
+ * @property {Float} expandBoundingBox - expand bounding box on all dimensions by this factor
+ *                               2 means double bbox, 3 means triple, etc.
  */
 
 export interface StageSignals {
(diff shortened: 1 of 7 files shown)


src/structure/data.ts
```diff
@@ -5,8 +5,8 @@ import { ValenceModel } from '../chemistry/valence-model'
 
 export interface Data {
   structure: Structure
-  '@spatialLookup': SpatialHash | undefined
-  '@valenceModel': ValenceModel | undefined
+  '@spatialLookup': SpatialHash | undefined | any
```
arose: Why the added `any`?

src/stage/stage.ts
```diff
@@ -102,6 +108,8 @@ declare global {
  * @property {Color} ambientColor - ambient light color
  * @property {Float} ambientIntensity - ambient light intensity
  * @property {Integer} hoverTimeout - timeout for hovering
+ * @property {Float} expandBoundingBox - expand bounding box on all dimensions by this factor
```
arose: I think `expandBoundingBoxFactor` would be a clearer name. Just  `expandBoundingBox` sounds to me like adding that many units to it.

arose:
Thanks @garyo!

src/structure/data.ts
garyo: Without those `| any`, I get this error:
```
> tsc && cpx "src/**/*.{vert,frag,glsl}" build/js/src/ && cpx "lib/*.js" build/js/lib/ && cpx "package.json" build/js/ && cpx "src/polyfills.js" build/js/src/

src/structure/data.ts(23,3): error TS2322: Type 'SpatialHash' is not assignable to type 'undefined'.
src/structure/data.ts(30,3): error TS2322: Type '{ charge: Int8Array; implicitH: Int8Array; totalH: Int8Array; idealGeometry: Int8Array; }' is not assignable to type 'undefined'.
npm ERR! code ELIFECYCLE
```

I doubt the `any` is the right solution, but it did work (build & run OK).

src/stage/stage.ts
garyo: I'll make that change. 
I didn't intend to push this commit into this PR, actually (thanks github), but now that it's here, let me explain the reason for it. I'm adding some Three.js objects into the scene inside `scene/rotationGroup/translationGroup`. They're bigger than the molecule. Without a change like this, they get fogged and clipped, because the bbox is set to the molecule size. (The Three.js objects are just straight-up Three.js geometry, not using the NGL object system.)
So if there's an alternate way to have the bbox include arbitrary other Three.js objects in the scene, that would be better than my change here. Any ideas appreciated!

garyo:
I'll redo this PR properly. Closing.

PR #599, opened 2018-11-12
Change Grid and AVHash to factory functions to fix #594

### LINKED ISSUE #594: Syntax error in minified blob when adding a surfacerepresentation
I have a clean checkout of the current `master` checkout, just pulled a few minutes ago.

```
git status
# On branch master
nothing to commit, working directory clean
```
The I build NGL with npm, and specifically the minified version:

    npm install
    npm run-script build-min

and then use  `dist/ngl.js`. When I create a surface representation:

     ngl_stage.compList[0].addRepresentation("surface")

I get no visible representation and the following error message in the browser javascript console:

    Object { stage: {…}, signals: {…}, parameters: {…}, uuid: "D3253E13-5F38-400A-9F03-7C48C5BFE0DC", parent: {…}, repr: {…} }
    Worker.onerror molsurf 
    error { target: Worker, isTrusted: true, message: "SyntaxError: function statement requires a name", filename: "blob:http://131.152.25.164:42342/53ef56ce-c48c-4929-8aa3-a9e0a7261fdb", lineno: 126, colno: 0, currentTarget: Worker, eventPhase: 2, bubbles: false, cancelable: true, … }
    ngl.js:1:691138
    SyntaxError: function statement requires a name
    53ef56ce-c48c-4929-8aa3-a9e0a7261fdb:126:8

Inspecting the blob, there indeed seems to be a syntax error. Here is an excerpt of the blob contents ([...] added for clarity):
```
[...]
function yS(C,I,c){var p=ty(I), [...]
function ym(t,e,i){t[0]=e[0]-i[0],t[1]=e[1]-i[1],t[2]=e[2]-i[2]}
function(C,I,L,R,t,e,i){var r=C.length,O=t[0],D=t[1],k=t[2],n=e[0][...]
self.func = function(t,e){var i=t.data.args,r=t.data.params;if(i&&r){var [...]
self.onmessage = function(t){var e=t.data.__name,i=t.data.__postId;void 0===e?console.error("message __name undefined"):void 0===self.func[...]
```
The third line is indeed missing the function name. This seems to be related with the minification process. If I use `build/js/ngl.dev.js` the surface representation is working properly.

One more note: if I checkout tag v2.0.0-dev.33 and run npm install && npm run-script build-min, surface representations work fine. Therefore this must be caused by a change since v2.0.0-dev.33.

fredludlow:
This fixes #594 for me (@xrobin could you see if it works for you?). `Grid` and `AVHash` are replaced with factory functions. This maintains the closure-style definition (rather than creating a class and setting prototype methods) which I think we found was helpful for performance back when adding the AVSurface code. That said, I don't have any benchmarks to back that statement up.

### DIFF
--- a/src/geometry/grid.ts
+++ b/src/geometry/grid.ts
@@ -5,21 +5,17 @@ import { NumberArray, TypedArray } from "../types";
  * @author Alexander Rose <alexander.rose@weirdbyte.de>
  * @private
  */
-interface Grid {
+export interface iGrid {
   data: TypedArray
   index: (x: number, y: number, z: number) => number
   set: (x: number, y: number, z: number, ...arg: number[]) => void
   toArray: (x: number, y: number, z: number, array?: NumberArray, offset?: number) => void
   fromArray: (x: number, y: number, z: number, array: NumberArray, offset?: number) => void
-  copy: (grid: Grid) => void
-  clone: () => void
-}
-export interface GridConstructor {
- (this: Grid, length: number, width: number, height: number, DataCtor: any, elemSize: number): void
- new (length: number, width: number, height: number, DataCtor: any, elemSize: number): Grid
+  copy: (grid: iGrid) => void
+  // clone: () => iGrid
 }
 
-const Grid = (function Grid (this: Grid, length: number, width: number, height: number, DataCtor: any, elemSize: number) {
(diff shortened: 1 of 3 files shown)


harryjubb:
Hey @fredludlow, I've had a similar problem to #594 with `Uncaught SyntaxError: Unexpected token (` in surface generation from the minified build, this branch fixes that :tada:

arose:
lgtm

Issue #602, opened 2018-12-01
NGL does not support some special formats of xplor/cns? 


hainm:
hi,

Please see the described issue here: https://github.com/arose/nglview/issues/761
The file is here: https://github.com/arose/nglview/issues/761#issuecomment-442341448

Thanks.

PR #603, opened 2018-12-10
Fix bug in xplor/cns parser, see #602

fredludlow:
I think this fixes bug originally reported in arose/nglview#761. As @shallerdavid suggested - it didn't handle files where the dimension of the map xy-sections was not a multiple of 6.


hainm:
thanks @fredludlow

arose:
Thanks @fredludlow!

PR #604, opened 2018-12-12
Bug in calculateBondsWithin

fredludlow:
There's a call to `ResidueType.getBondIndex` that uses the atomID from the structure (including the offset) as opposed to the appropriate one for the ResidueType


arose:
ouch, thanks @fredludlow

PR #605, opened 2018-12-12
Marching-cubes bug

fredludlow:
Deep in the marching cubes algorithm, when interpolating between grid points and generating a new vertex, that vertex gets cached/looked up (as you'll be generating the same point when processing multiple cubes). However, the vertex is stored based only on the origin grid point not "gridpoint plus direction", so some vertices are in the wrong place and some triangles end up having duplicate vertices (i.e. the vetex indices of a triangle might be 1,1 and 2)

I fixed this for contoured surfaces some time ago, but for some reason didn't apply the fix to solid/wireframe surfaces, I guess I thought it required a larger array and I couldn't actually see anything wrong visually with the old version. Anyway - I've now found some examples where this is really definitely wrong - this simplifies the code a bit and removes the bug.

### DIFF
--- a/src/surface/marching-cubes.ts
+++ b/src/surface/marching-cubes.ts
@@ -397,12 +397,9 @@ function MarchingCubes (this: MarchingCubes, field: number[], nx: number, ny: nu
       normalCache = new Float32Array(n * 3)
     }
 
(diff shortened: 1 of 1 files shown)


arose:
thanks @fredludlow

PR #609, opened 2018-12-21
Add toJSON to buffer.ts to avoid circular refs when exporting

garyo:
The groups' children refer back to the mesh, and the picking object
has a circular structure.data attribute.

I don't believe the groups or picking are required for
serialization (e.g. model export) so this change shouldn't impact
normal use.


arose:
I am ok with this, @ppillot and @fredludlow any comments?

ppillot:
Looks fine to me. Maybe the typing for the result variable could be `Partial<Buffer>` instead of `any`

fredludlow:
Okay with me - merged

arose:
thanks!

PR #613, opened 2019-01-10
Add expandBoundingBoxFactor method to stage

garyo:
This can be useful if NGL is embedded and additional graphics objects
are added to the stage outside NGL. Without expanding the bounding
box, those additional objects may be clipped or fogged.

This is the improved version of this patch, with better math
and also it only applies the expansion once.

### DIFF
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -102,6 +102,8 @@ declare global {
  * @property {Color} ambientColor - ambient light color
  * @property {Float} ambientIntensity - ambient light intensity
  * @property {Integer} hoverTimeout - timeout for hovering
+ * @property {Float} expandBoundingBoxFactor - expand bounding box on all dimensions by this factor
+ *                               2 means double bbox, 3 means triple, etc.
  */
 
 export interface StageSignals {
@@ -138,6 +140,7 @@ export const StageDefaultParameters = {
   ambientIntensity: 0.2,
   hoverTimeout: 0,
   tooltip: true,
+  expandBoundingBoxFactor: 1,
   mousePreset: 'default' as MouseControlPreset
 }
 export type StageParameters = typeof StageDefaultParameters
@@ -263,6 +266,7 @@ class Stage {
     viewer.setSampling(tp.sampleLevel)
     viewer.setBackground(tp.backgroundColor)
     viewer.setLight(tp.lightColor, tp.lightIntensity, tp.ambientColor, tp.ambientIntensity)
+    viewer.setExpandBoundingBoxByFactor(tp.expandBoundingBoxFactor)
 
     this.signals.parametersChanged.dispatch(this.getParameters())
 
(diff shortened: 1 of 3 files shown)


garyo:
Any interest in merging my new fixed version of this? I believe it addresses all the issues you mentioned in the previous PR.

fredludlow:
So I get slightly odd behaviour with this:
```javascript
stage.setParameters({expandBoundingBox: 2.0})

// Load something small
stage.autoView() 
// Looks right (about half the size of normal)

// Load something much bigger/different place (I guess this updates the bouding box)
stage.autoView()
// Seems to have forgotten expandBoundingBox setting

stage.setParameters({expandBoundingBox: 2})
stage.autoView() // Now it's working...
```

Maybe this doesn't matter as 
1) it isn't what you were trying to achieve? (I don't have an example to play with which has other things in the scene).
2) I've misunderstood how I'm supposed to use it

garyo:
Hi Fred; thanks for the review. I'm still figuring out the details of how NGL works with Three. In my app, I have an NGL object (the molecule) as well as some non-NGL Three.js objects in the scene. (They come from an external 3d package so don't really fit into the NGL representation scheme.) 
The reason for this PR is NGL always sets the Z clipping planes and fog based on its computed bounding box, which causes most of my external Three.js objects to be clipped or fogged. To be honest I'm not saying this is the only or best way to do this, but it is working for me for now.
And yes, you're right that if you load something different it doesn't reapply the expansion -- when I didn't reset the expansion, it would happen repeatedly (2x, then 4x, then 8x etc.) so I make sure to only do it once. If there were an explanation of the NGL render pipeline I could read, that might help me do this better. For instance maybe instead of a factor, I should simply union in a passed-in bbox?
Or even more simply, if I could just set the Z +/- limits and turn off fogging that would also work for me I think. (I started on that but couldn't figure out where to make the changes -- any help would be appreciated!)

fredludlow:
Sorry for the delay again... I've been testing this in house here and had to revert it due to weird behaviour on large structures (clipNear/clipFar at 0/100 wouldn't show the whole protein).

The clipping parameters are currently relative, which is a bit of a pain - I have a method for serializing views but the actual displayed clipping depends on what else is loaded (and not displayed). 

So, maybe we could make clipNear and clipFar take absolute values - the mouse shortcut for changing clipping could be bounded at a reasonable upper limit based on some factor of boxSize to avoid the situation where you end up setting them to extreme values and can't work out how to get back? Would need to have a think about what if anything this would break...

garyo:
Hi Fred; thanks for your comments. I'm definitely interested in pursuing this. You're right that my ultimate concern is with the clipping limits and fog, not really bounding box. 
A way to set clipping limits (and fog) in absolute coords would be really helpful. If that's not practical (for compatibility reasons or whatever), perhaps we could add an optional callback to update them after they've been set internally? `onUpdateClipLimits` and `onUpdateFogLimits` or something like that?

garyo:
I'm closing this PR -- hopefully #656 (absolute clip/fog) is seen as a better approach.

PR #635, opened 2019-03-20
Surface tweaks

fredludlow:
A few fixes for surfaces and electrostatic coloring

Specifically, fix the buggy surfaces seen at high scaleFactor values in 'av' mode, assign correct atom indices, and tidy up electrostatic coloring code.


arose:
awesome, thank you

PR #637, opened 2019-03-27
tiledrenderer gap fix

martingraham:
ok, following (sort of) fredludlows advice, I made a clean install of ngl and just changed/committed tiledrenderer.ts to avoid all the cruft that had accumulated in my version


fredludlow:
Thanks @martingraham! 
So with your version I can't find any set of parameters that reproduce the bug. I also can't reproduce with dev.35..? 

Merging anyway as the newer version is clearer (to me at least)

PR #654, opened 2019-05-16
Fix nucleic saccharide

### LINKED ISSUE #650: LAT (lactose) in PDB files (e.g. 1JYN) wrongly typed as nucleic instead of saccharide
This issue does not happen with mmtf files as the residue detection can be based on the chemcomp type.
When this information is absent (in PDB files), saccharide detection in the `isSaccharide()` method relies on residues names. But `isRna()` detection relies on atoms identifiers and the combination O3', C4', O2' suffices to make a residue of RNA type (see https://github.com/arose/ngl/blob/91b32813c794e0fe7248021977d662638d303638/src/store/residue-type.ts#L209).
LAT has the pattern O2', O3', C4' in its [nomenclature](http://ligand-expo.rcsb.org/reports/L/LAT/index.html): isRna() returns true with this residue.
As isRna() is searched before isSaccharide() in the [getMoleculeType()](https://github.com/arose/ngl/blob/91b32813c794e0fe7248021977d662638d303638/src/store/residue-type.ts#L143) method, LAT is then incorrectly set to RNA.

I don't know what the best fix would be for this issue. Is the isRna() test stringent enough? Should the isSaccharide() test be moved before (as it requires testing against a long list of residues names for residues that are less frequent, it would make the structure building as a whole less efficient)?

ppillot:
This branch fixes #650 
It adds a check for hetero value in the `isDna()` and `isRna()` methods before testing for specific atoms value such as O3' (which are also present in disaccharides).

It contains also a second commit for adding I and DI as base names in conformity with what Jmol does.

### DIFF
--- a/src/store/residue-type.ts
+++ b/src/store/residue-type.ts
@@ -203,6 +203,8 @@ export default class ResidueType {
   isRna () {
     if (this.chemCompType) {
       return ChemCompRna.includes(this.chemCompType)
+    } else if (this.hetero === 1) {
+      return false
     } else {
       return (
         this.hasAtomWithName(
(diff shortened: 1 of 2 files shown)


fredludlow:
Thanks @ppillot

PR #656, opened 2019-05-30
Allow clip and fog to be specified as absolute

garyo:
Per discussion in #613, this is a hopefully better approach to the same problem.

The problem is how to handle clipping and fog when the scene contains not just the NGL objects, but other three.js objects. NGL doesn't know about them, so sets clipping and fog based on the bounding box of its own scene. Previously in #613 I suggested expanding the bbox by a constant factor, which is problematic for various reasons (see that PR). In there, @fredludlow suggested a different approach: allow clipping and fog to be specified in absolute units, rather than relative to the bbox size. This PR implements that.

It preserves all existing behavior; the new absolute behavior is optional, based on stage params `clipIsAbsolute` and `fogIsAbsolute` and new optional args to `setClip` and `setFog`. When those are set, the clip near/far and fog near/far are interpreted as absolute units. (I changed them from ints to floats so fractional units work.)

As for testing, it works fine for me, but more eyes would certainly be helpful.


fredludlow:
Thanks for this @garyo, I'll try and test it today.

fredludlow:
I think this is going along the right lines, two thoughts:

1) If you use orthographic mode and zoom out then objects disappear (I haven't had time to dig yet, but changing zoom calls __updateClipping I think, which does some necessary stuff)
2) Should we make absolute the default? (should definitely check what @arose thinks - Alex?)

garyo:
Hi Fred -- let me know what you learn re: ortho. (Is there any test suite for NGL that I should add tests to?)

As for making it the default, either's fine with me (I'll always use absolute) but I figured when in doubt, be back-compatible -- so I didn't do that in my PR.

fredludlow:
Right, nearly there now I think:

The problem isn't just with ortho, see my checkout of your PR here:
https://www.fredludlow.com/ngl/abs-clip-garyo/examples/webapp.html?script=color/bfactor

If you set clipIsAbsolute to true and zoom out the protein disappears - this is because zooming is done by moving the camera z-position, and the camera clipping planes need to take into account this changed distance. The ortho camera is also moved (its z-position is irrelevant but I guess it's easier to keep the two in sync). 

I think I've got it working here by adding 'cDist' value:
https://github.com/garyo/ngl/compare/feature/clip-fog-abs...fredludlow:pr-656?expand=1#diff-c2341ad21e7bbd05a2082e443473d3eaL1059

see:
https://www.fredludlow.com/ngl/abs-clip-zoom-fix/examples/webapp.html?script=test/clipping

I'll make a PR to your branch so the whole thing still sits under this PR

There's a couple of other things that would be good to get ironed out before merging into master:

* Remove debugClipping from the code if it's unused
* The behaviour of clipNear=0 in absolute clips out everything in front of the centre, but in relative mode it means show everything in front of the centre, I'm not sure how to unify these nicely? I think this blocks making absolute the default so if we can't come up with a nice way of doing it let's leave the default as relative as per your last comment.

garyo:
I'm working on this today -- thanks for adding the really nice example!

My original intent was indeed to ignore `cDist` in absolute mode, so the near and far are purely camera-based. The thinking being that the model isn't the only thing in the three.js scene, so the user would specify near and far in the more traditional OpenGL style.

Your patch is a compromise: it still has `clipNear` and `clipFar` being distances rather than scale factors, but they're relative to the world-space origin (if I understand correctly, that's what adding `cDist` does -- turn them from camera-relative to origin-relative). I haven't seen a graphics system that works that way, but I see how it could be useful for a scene centered around the origin.

By the way, your change switches clipNear to go the other way from my change; I was expecting them both to be `+`. I think in absolute mode `+` should always be further from the camera, and `-` closer, like OpenGL (and like relative mode).
```typescript
if (p.clipIsAbsolute) {
    this.camera.near = this.cDist - p.clipNear // I prefer + here
    this.camera.far = this.cDist + p.clipFar
 ...
```

So, getting to the point... since your version has 0 meaning world origin, setting either `near` or `far` to 0 won't be super useful as a default. I suspect the right answer here is absolute just shouldn't be the default -- there's no way to set it "right" for all cases. In my version you _could_ have absolute as the default if you set `near=0.001` and `far=100000` or something, but that has disadvantages as well, such as reduced z-buffer accuracy vs. a tighter bound.

One other thought: won't making `near` origin-based make the actual camera clipping plane be negative when the camera is inside the model? I'm pretty sure negative zNear is disallowed in OpenGL. (https://www.khronos.org/opengl/wiki/Depth_Buffer_and_Perspective_Rendering) -- maybe something elsewhere in the pipeline clamps it though.

After all, I think I still prefer my original version -- but I can work with an origin-based setting as well. Whatever you folks think is best.

fredludlow:
> Your patch is a compromise: it still has clipNear and clipFar being distances rather than scale factors, but they're relative to the world-space origin (if I understand correctly, that's what adding cDist does -- turn them from camera-relative to origin-relative). I haven't seen a graphics system that works that way, but I see how it could be useful for a scene centered around the origin.

Right, so I guess all the molecule viewers I've used follow this convention - rotation is centred on a point and "zooming" in pymol, astexviewer, Maestro and (current) NGL is moving the camera (perspective mode) or changing the camera size (orthographic) but keeps clipping planes relative to the centre of rotation.

I think I hadn't understood what you're trying to achieve - am I right in saying you'd like the zoom controls to do something akin to the forwards/backwards keys in a fps game? 

It seems there are two actions getting bundled together here. There's an action that moves the focal point in the z-direction. (this is missing from NGL at the moment), and then there's the action for changing the scale but keeping the focus the same (which is named "zoom" in NGL but is implemented with a camera z-move, presumably to avoid extreme zoom values for the perspective camera).

It would be really nice to have both actions available (which I think is independent of whether clip is expressed in absolute terms or relative). So how about (thinking aloud here...):

We create new MouseActions "zPanDrag" and "zPanScroll" which move the focal point forwards and backwards (they'll end up calling `Stage.trackballControls.pan(0,0,z)`), the camera position would also get updated automatically. We could create a new mouse profile where this is bound to the 'scroll' event specifically for your use-case.

Then you'd need to initialize your stage with: {clipIsAbsolute: true, fogIsAbsolute: true, mousePreset: 'insertSensibleName'}. Your clip values would still be "Angstroms from the centre of the scene" but instead of just changing camera you'd be changing both the focal point of the scene and the camera. It's a bit convoluted but I think the behaviour would be similar to your original PR? Then for "normal" usage we can have absolute clip values and another key-mouse combination bound to a focal point move.

> One other thought: won't making near origin-based make the actual camera clipping plane be negative when the camera is inside the model? I'm pretty sure negative zNear is disallowed in OpenGL. (https://www.khronos.org/opengl/wiki/Depth_Buffer_and_Perspective_Rendering) -- maybe something elsewhere in the pipeline clamps it though.

I think it does get clamped somewhere, but yes, we need to make sure zNear stays positive!



Anyway, sorry this has dragged on a bit, let me know if I'm finally getting the right end of the stick or not!

garyo:
Hi Fred -- it's not so much that I'm interested in particular mouse/camera controls (though that is certainly interesting in the larger sense). All I really want to do is turn off clipping and fog altogether, so my whole scene is shown; not just the molecule that NGL knows about, but all the other three.js objects I'm adding to the scene.

In most graphics systems including WebGL, the near and far clipping planes set the range of the z buffer, which is why I think they're usually camera-relative; see OpenGL or [DirectX](https://docs.microsoft.com/en-us/windows/desktop/direct3d9/viewports-and-clipping) - they're a vital part of the perspective transform. I can see the value of NGL dynamically changing them when the camera moves so they appear to be scene-relative; it keeps the z buffer tightly around the scene no matter where the camera goes. But in my case, NGL's molecule isn't the whole scene, so it can't really know how to set them appropriately.

My quibble with your version, I guess, is that 0 isn't a particularly useful value in absolute mode, since it means "origin" i.e. the middle of the molecule. So If we go with your version (adding `cDist`) I'll just set the clip & fog values to -10000 and +10000 or something like that, and assume the near will get clipped to the camera. We can just expect that anyone like me using absolute mode will set the values appropriately for their scene. 

A secondary quibble is it seems better to decouple things when possible. Coupling absolute clipping planes to camera motion means any extension to NGL to do fancier camera tracks (as you suggest in your comment above, and as I will want to do in the future) has to handle resetting the absolute clipping planes on every frame. Just something to be aware of I guess.

Big picture, I guess I'm uncertain what the intent of having the clipping planes is in NGL -- is it intended for visualizing interior structure, or is it more like a traditional graphics system, to optimize performance (cull objects behind the camera and too far away to be seen) and set the Z buffer range (to avoid zebra striping for overlapping objects)?

That raises the question of using clipping planes for viz purposes, to slice the object. I actually don't think the current clip planes are the best way to do that -- three.js now has arbitrary world-space clipping planes (see https://threejs.org/examples/webgl_clipping.html for example) which are really nice for this. Would be a nice extension to NGL!

fredludlow:
> Big picture, I guess I'm uncertain what the intent of having the clipping planes is in NGL -- is it intended for visualizing interior structure, or is it more like a traditional graphics system, to optimize performance (cull objects behind the camera and too far away to be seen) and set the Z buffer range (to avoid zebra striping for overlapping objects)?

For use in a drug discovery context (and I'm biased here as this is my field) it's predominantly for visualization - you're frequently looking at ligands bound in pockets in the protein with all the atoms, bonds, possibly surfaces, interactions displayed and you want to cut out the "noise" of things far in front/behind the ligand.

> That raises the question of using clipping planes for viz purposes, to slice the object. I actually don't think the current clip planes are the best way to do that -- three.js now has arbitrary world-space clipping planes (see https://threejs.org/examples/webgl_clipping.html for example) which are really nice for this. Would be a nice extension to NGL!

I'll take a look at those - though I suspect the maths would be more complex (you'd end up needing to move them every time the camera moved?) - though being able to set them independently of camera orientation would be interesting (and a potential UI/UX nightmare!)

Getting back to our particular issue. I agree with you that my patch to your PR isn't the right way to do this (at the time I'd misunderstood what you were trying to achieve).

How about:
`clipMode`: `'scene'|'camera'` Where `scene` is the current behaviour and camera is the behaviour you originally proposed (`clipNear and `clipFar` get passed straight to the camera and the dev is responsible for giving them sensible values).
`clipScale`:`'absolute'|'relative'` (whether or not values are intepreted are a percentage of boxsize or not, only really applicable to clipMode `scene`)

This is basically your original PR! (sorry) but with clipIsAbsolute/fogIsAbsolute replaced by a clipMode setting. You might be able to think of a better name than clipMode but having 'absolute' in there was what lead me off track in the first place. 

The stuff about "z-move" actions etc we can handle somewhere else when it becomes a priority for somebody.

As always, if I'm still barking up the wrong end of the stick do say ;)

garyo:
This makes total sense, Fred -- thanks for sticking with this. 
I'll redo my patch with `clipMode` and `clipScale` -- presumably also `fogMode` and `fogScale`. 
As you note, `clipScale` will be ignored in `scene` mode. That leaves the only questionable case being `clipMode=camera; clipScale=relative` but I'll implement that (clip values are camera-relative and in units of percent-of-NGL-scene-size, 0 still means right at the camera plane) and if anyone wants it, it'll be there. I will of course leave the defaults as `clipMode=scene; clipScale=relative`. (Same for fog throughout.)

As for world-space clipping planes, check out the sample at the link I sent. It's pretty nice. It has a plane-slice UI slider for each clipping plane (but no rotation). Since the clipping plane is spec'ed in world coords, you don't have to do anything when the camera moves. It just slices the object (either the whole world, or a particular object -- both are supported).
Yes, it would need a UI for moving and rotating the clipping plane. Probably a rotate-sphere and a single translate slider (along the normal) like the sample I linked would suffice. But as you say, that's a whole other (fun!) discussion.

garyo:
OK @fredludlow check out this version, I think it covers our discussion above.

fredludlow:
Great, looks like we're basically there. Comments:

1) I still prefer keeping clipNear and clipFar both as positive parameters in scene/absolute mode. I'm hoping that's not controversial as it's not a mode you'll be using I guess? My reasoning is it keeps some of the GUI behaviour simpler (a range from 0-200 then covers pretty much all needs for abs and relative)
2) The behaviour of Shift-scroll in scene mode should be consistent between abs and relative
3) The behaviour of Shift-scroll in camera mode may or may not need fixing, depending on what you'd like to do with it?
4) In the default webapp GUI, there's a "Settings" widget and a "Preferences" widget. Do we want to expose any of these other modes in "Settings"? (My guess is no as nobody else has commented on the relative/absolute clipping or camera/scene basis - it's an issue when building larger applications, and if you're doing that you've got the API)

I'm happy to address 1) and 2) in a separate PR

What are your thoughts on 3 and 4?

garyo:
1: makes sense, I can flip that easily. I'll update the PR.
2 & 3: aha, I didn't know about shift-scroll (mouse action `focusScroll`, right?). Also `clipNearScroll` looks like it might need a tweak, though it's not clear to me that that ever gets called (looking at `clipNearScroll` though, maybe it's fine as is?).
I agree shift-scroll should behave consistently in scene mode between absolute & relative. Is the right place to do that in stage.ts, `setFocus()`? Perhaps better for you to take that on; I don't totally understand the intent of shift-scroll/focusScroll/setFocus. Are they supposed to pull the clipping planes in toward the center of the model? The comment just says "Move focus planes based on scroll-delta" so I'm not sure.
In camera mode I don't think it makes sense to do this kind of thing -- I'd use three.js webgl world-space clipping in my app. I'm happy if we just disable driving clipping planes from the mouse in that mode. 
And for your #4, I agree -- camera mode is probably only useful for app developers. I doubt a direct end user would ever want to enable it.
So once I update to fix the clipNear sense (did it just now), can we consider this PR mergeable, then you can look into focusScroll behavior? I'll be happy to help of course.

fredludlow:
> Are they supposed to pull the clipping planes in toward the center of the model?

Exactly this.

PR #663, opened 2019-06-12
Avoid more circular refs when exporting

garyo:
Same fix as #609, applied to DoubleSidedBuffer since that does not extend Buffer.

I actually don't know if _any_ of this three.js userData is useful when exporting/serializing, but seems better to be safe and export whatever might be useful.


fredludlow:
Thanks!

> I actually don't know if any of this three.js userData is useful when exporting/serializing, but seems better to be safe and export whatever might be useful.

Me neither :)

PR #665, opened 2019-06-18
Be able to select atoms and move them around

lilyminium:
This pull request adds the following:

- Being able to select or deselect atoms with shift+click, which is indicated by a transparent white VdW structure
- Being able to deselect all atoms with shift+double click
- Being able to drag selected atoms with shift+right-drag to alter their coordinates

There is room for other transformations, e.g. rotation (which I started on but isn't working very well). Being able to select by clicking is very convenient.


fredludlow:
Thanks @lilyminium, this sounds great! I'll have a proper look at it as soon as I can.

ppillot (changes requested):
Thank-you for making this PR, which would provide a good starting point for adding some editing capabilities to NGL, a much demanded new set of features!

I am really sorry to start the review process almost 4 years after you had submitted the PR in the first place. I understand that since then, you might have moved to other subjects and might not want to spend more time on this feature. 
Nonetheless, given the implications of adding this kind of capabilities to NGL, I think we should proceed cautiously.
I can see that there have been several PR made in the direction of modifying the coordinates of atoms (one by @fredludlow and one by @garyo are still pending).

From the code review, I also noticed that there is a need for some utility functions in the NGL core, or some improvements to existing functions parameters (such as the way selections could be passed as strings, selection objects or atom sets interchangeably).

Based on that, a possible roadmap would be:
- I make some changes to NGL to make the necessary modifications to pass atom sets as selections to the various utility functions (repr.setSelection, structure.eachAtom,...)
- In a first step, this PR can be divided to add the capability to make interactive selections
- During that time, it would be nice to have @fredludlow 's and @garyo 's PR addressed.
- In a second step, we make a second PR to add the capabilities to translate atoms from the selection

lilyminium:
Thanks for the thoughtful and comprehensive review, @ppillot! The roadmap you suggested looks great and practical. Unfortunately it has been a while since I looked at NGLViewer or Javascript in general, so it would take some time for me to get my head back into this. I don't want to hold up any plans -- would it be better to close this PR and eventually work towards the two divided functionalities?

ppillot:
No worries! I'll leave it open for now as it's a useful starting point.
I've opened the discussion with @fredludlow and we'll see how we advance starting from there.
Since then, I've already started working on more flexible selections which will accept BitArrays as their definitions. Once this is in, we'll hopefully be in a better position for the next steps.

PR #673, opened 2019-07-02
FIX fullscreen icon

hainm:
Otherwise:

Screen Shot 2019-07-02 at 7 39 20 PM

PS: May be this happens due to adding new camera stereo type?


fredludlow:
Thanks @hainm!

Issue #671, opened 2019-07-01
Fix focusScroll behaviour when clipMode:scene clipScale: absolute


fredludlow:
See discussion in #656

fredludlow:
Fixed in #674 I think

PR #674, opened 2019-07-08
Consistent focusScroll behaviour between relative/absolute clipping

fredludlow:
See #671 and #656 

As discussed, this ensures focusScroll behaves sensibly for both relative and absolute clipScale.

@garyo  I'd be grateful for any feedback on this one. One significant change I made was removing the fogMode and fogScale parameters (they now just follow clipMode and clipScale). I thought it made the logic quite complex for what I think would be a very unusual use case (i.e. camera clipping, scene fogging)?

### DIFF
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -90,8 +90,8 @@ declare global {
  * @property {Float} clipFar - position of camera far/back clipping plane
  *                               in percent of scene bounding box
  * @property {Float} clipDist - camera clipping distance in Angstrom
- * @property {String} clipMode - how to interpret clipNear/Far values: "scene" for scene-relative, "camera" for camera-relative
- * @property {String} clipScale - "relative" or "absolute": interpret clipNear/Far as percentage of bounding box or absolute Angstroms (ignored when clipMode==camera)
+ * @property {String} clipMode - how to interpret clipNear/Far and fogNear/Far values: "scene" for scene-relative, "camera" for camera-relative
+ * @property {String} clipScale - "relative" or "absolute": interpret clipNear/Far and fogNear/Far as percentage of bounding box or absolute Angstroms (ignored when clipMode==camera)
  * @property {Float} fogNear - position of the start of the fog effect
  *                               in percent of scene bounding box
  * @property {Float} fogFar - position where the fog is in full effect
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -96,8 +96,6 @@ declare global {
  *                               in percent of scene bounding box
  * @property {Float} fogFar - position where the fog is in full effect
  *                              in percent of scene bounding box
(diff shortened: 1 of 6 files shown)


src/stage/stage.ts
```diff
@@ -96,8 +96,6 @@ declare global {
  *                               in percent of scene bounding box
  * @property {Float} fogFar - position where the fog is in full effect
  *                              in percent of scene bounding box
- * @property {String} fogMode - how to interpret fogNear/Far values: "scene" for scene-relative, "camera" for camera-relative
```
garyo: I think this totally makes sense to remove these; I can't imagine anyone wanting clip one way and fog the other. The docs above should be adjusted to note that clipMode/clipScale also affect fog.

src/viewer/viewer.ts
```diff
@@ -1032,74 +1025,78 @@ export default class Viewer {
...
+
+      if (p.clipScale === 'absolute') {
+
         this.camera.near = this.cDist - p.clipNear
         this.camera.far = this.cDist + p.clipFar
+        fog.near = this.cDist - p.fogNear
+        fog.far = this.cDist + p.fogFar
+
       } else {
-        // relative scene mode (default): convert percentages to Angstroms
```
garyo: I liked those comments, but OK

src/viewer/viewer.ts
```diff
@@ -1032,74 +1025,78 @@ export default class Viewer {
...
     const p = this.parameters
 
-    const debugClipping = false
+    // bRadius must always be updated for material-based clipping
+    // and for focus calculations
+    this.bRadius = Math.max(10, this.boundingBoxLength * 0.5)
 
-    // clipping
+    // FL: Removed below, but leaving commented as I don't understand intention
+    // this.bRadius += this.boundingBox.getCenter(this.distVector).length()
```
garyo: I don't know about this enlarging the bbox either -- I did a bit of git archaeology; it blames back at least to 600dc160b (@arose 2016-06-22, "refactored/cleaned-up camera/view controls"):
```diff
-        bRadius += boundingBox.center( distVector )
-            .add( rotationGroup.position )
-            .length();
+        bRadius += boundingBox.center( distVector ).length();
```

src/viewer/viewer.ts
fredludlow: Sorry - they can go back! (It was the result of several refactors rather than a conscious decision to remove them!)

src/stage/stage.ts
fredludlow: Thanks, will update docs

src/viewer/viewer.ts
fredludlow: They're back :)

src/viewer/viewer.ts
fredludlow: Hmm, perhaps Alex remembers? @arose

garyo:
LGTM!

garyo:
@arose do you approve this PR?

PR #682, opened 2019-07-26
Updates to kinemage parser

vbchen:
Fixed some bugs with the kinemage parsing code.
Added some initial testing code using Jest to check for correct parsing of kinemages.


fredludlow:
I also don't know much about the file format, but if it's working then great!

Minor comment (mostly on me for not looking at it earlier) but for older PRs like this is it better to merge as is, or create a new branch, cherry pick the commits (preserving committer/author details) and create a cleaner PR? Pros are a cleaner git history and getting rid of unecessary rebase/merge comits, cons are it requires two maintainers rather than one as one of us needs to approve the other's new PR.
(To be clear, I'm completely happy to merge this now for expediency's sake!)

garyo:
@fredludlow I'm also generally in favor of a cleaner history. I merged this now since it's been pending for a long time. It's a bit complicated to rebase while preserving the author details (especially committer), but it can be done in most simple cases. I'm happy to try that going forward.

PR #698, opened 2019-10-06
performance optimisation when stage has been disposed of

aurel-l:
I realised that there were some functions still being called recursively at every frame even after removing the scene.
In hope of getting rid of as much as I could, I was removing everything by doing:
```js
stage.removeAllComponents();
stage.dispose();
stage.viewer.container.innerHTML = "";
stage = null;
```
This might not be too much of a problem in some pages, but if you happen to add and remove stages a lot (popup/modal with ngl viewer, or single-page applications) then this can build up.

Screenshot from 2019-10-06 22-40-25
After adding and removing a stage 20 times (using a modified version of `/examples/embedded.html`) This is what I would get on the performance tab of Chrome.
Every single frame I would have 20 calls to `mouseObserver._listen()` and 20 to `viewer.animate()` that would amount to about 5ms of useless use of the main thread, and memory usage growing (even if it does get caught by the garbage collector eventually).

Compare to this screenshot using the code in this PR
Screenshot from 2019-10-06 22-42-14
Here with have absolutely nothing happening after doing the same steps, which is way better regarding resource usage.

### DIFF
--- a/src/stage/mouse-observer.ts
+++ b/src/stage/mouse-observer.ts
@@ -142,6 +142,7 @@ class MouseObserver {
   lastClicked: number
   overElement: boolean
   lastTouchDistance: number
+  private frameRequest: number
 
   /**
    * @param  {Element} domElement - the dom element to observe mouse events in
@@ -218,7 +219,7 @@ class MouseObserver {
         this.signals.hovered.dispatch(cp.x, cp.y)
       }
     }
-    window.requestAnimationFrame(this._listen)
+    this.frameRequest = window.requestAnimationFrame(this._listen)
   }
 
   /**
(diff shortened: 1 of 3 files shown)


aurel-l:
OK, so I had installed this LGTM analysis thing on my forked repo and force-pushed to retrigger a check but I don't think it's something that needs to be done from my side. Let me know if I have to do something in order to pass the check

garyo:
This looks basically OK, though I'm not expert on this code, but perhaps you should protect against accidentally calling `window.cancelAnimationFrame` with an undefined value, and set the request back to undefined once the request is canceled. (And type `frameRequest` to include undefined.)

sto3psl:
> but perhaps you should protect against accidentally calling `window.cancelAnimationFrame`

This is not strictly necessary. `cancelAnimationFrame` does nothing when called with `undefined` or any invalid value.

ppillot (approved):

ppillot:
Reviewed and tested after a rebase. This works as advertised. Thanks @aurel-l  for your contribution!

PR #716, opened 2019-12-11
Update to three.js r111

garyo:
This was done in stages to keep things clean and separated. First r103, then r106, then r111.
111 is the most complicated because it requires updating to typescript 3.7.2, which showed over 100 new errors. Most of them were just type issues, but a few were real potential bugs.
A couple are just annoying, but needed, in buffer.ts for instance:
```diff
-      m[ name ] = value
-      wm[ name ] = value
-      pm[ name ] = value
+      (m[ name ] as any) = value;
+      (wm[ name ] as any) = value;
+      (pm[ name ] as any) = value
```
because typescript now has problems with assigning to boolean class members indirectly. (Apparently it was never type-safe in the old versions anyway, so this is "more explicit", sigh.)

Hopefully the rest of the changes should speak for themselves.


src/viewer/viewer.ts
```diff
@@ -483,7 +483,8 @@ export default class Viewer {
     // workaround to reset the gl state after using testTextureSupport
     // fixes some bug where nothing is rendered to the canvas
     // when animations are started on page load
-    this.renderer.clearTarget(this.pickingTarget, true, true, true)
```
ppillot: Could you explain why you made this change? Apparently clearTarget() is still a valid method of renderer, I am wondering if there is an advantage?

src/viewer/viewer.ts
```diff
@@ -1148,9 +1149,11 @@ export default class Viewer {
   }
 
   private __renderPickingGroup (camera: PerspectiveCamera|OrthographicCamera) {
-    this.renderer.clearTarget(this.pickingTarget, true, true, true)
+    this.renderer.setRenderTarget(this.pickingTarget || null!)
```
ppillot: It looks like the `!` indicator can now be removed as `null` is an accepted argument in three.js typings. Same applies for l:1156 below

src/surface/volume.ts
```diff
@@ -293,7 +293,7 @@ class Volume {
 
     const box = this._getBox(center, size)
     const sd = this.volsurf.getSurface(
-      isolevel, smooth, box!, this.matrix.elements, contour, wrap
+      isolevel, smooth, box!, new Float32Array(this.matrix.elements), contour, wrap
```
ppillot: I guess you are passing `matrix.elements` to the Float32Array constructor here to avoid TS complaints. There is a perf penalty in that statement. As it does not change the way the code runs, I'd suggest forcing the typing here with a `this.matrix.elements as unknown as Float32Array`
Same goes for l:382

src/buffer/geometry-buffer.ts
```diff
@@ -152,14 +152,16 @@ abstract class GeometryBuffer extends MeshBuffer {
           position[ i3 ], position[ i3 + 1 ], position[ i3 + 2 ]
         )
         this.applyPositionTransform(matrix, i, i3)
-        applyMatrix4toVector3array(matrix.elements, transformedGeoPosition)
```
ppillot: Same as in surface/volume.ts: force the typing here with a `this.matrix.elements as unknown as Float32Array`
(and same l:163)

src/align/superposition.ts
```diff
@@ -203,7 +203,7 @@ class Superposition {
     // do transform
 
     const mult = new Matrix(4,4)
-    mult.data = transform.elements
+    mult.data = new Float32Array(transform.elements)
```
ppillot: Same comment as in surface/volume.ts

src/align/superposition.ts
```diff
@@ -133,7 +133,7 @@ class Superposition {
     multiplyABt(tmp_1,add,tmp_2)
 
     transpose(transformMat_,tmp_1)
-    this.transformationMatrix.elements = transformMat_.data
+    this.transformationMatrix.elements = Array.from(transformMat_.data)
```
ppillot: If this is made only for avoiding TS complaining, I would suggest to replace this operation by a forced type casting: `transformMat_.data as unknown as number[]`

src/component/structure-component.ts
```diff
@@ -255,7 +255,7 @@ class StructureComponent extends Component {
 
   addRepresentation <K extends keyof StructureRepresentationParametersMap>(
     type: K,
-    params: Partial<StructureRepresentationParametersMap[K]> = {},
+    params: Partial<StructureRepresentationParametersMap[K]>|{defaultAssembly: any} = {},
```
ppillot: This should be added to structure-representation.ts, l:38 and typed as a `string` (not `any`)

src/stage/stage.ts
```diff
@@ -600,7 +600,7 @@ class Stage {
    */
   toggleFullscreen (element: HTMLElement) {
     if (!document.fullscreenEnabled && !document.mozFullScreenEnabled &&
-        !document.webkitFullscreenEnabled && !document.msFullscreenEnabled
+        !(document as any).webkitFullscreenEnabled && !document.msFullscreenEnabled
```
ppillot: it's a matter of taste there, but one can also legitimately use `//@ts-ignore` before the non standard statements

src/surface/volume.ts
garyo: Good point about performance. But I don't think your cast idea will work, because `this.matrix.elements` is actually an array of doubles ("number[]"), so you can't just cast it to `Float32Array`, right? It would look like garbage. And a `matrix` is pretty small, so the copy/convert overhead shouldn't be *that* bad.

garyo: Well, I wrote this whole comment about how that couldn't possibly work because `matrix` is a `number[]` so it's doubles, and `Float32Array` is floats, so the cast should come out as garbage (can you tell I'm an old-school C/C++ guy?) but I did a simple standalone test and it does work. By "work" I mean the generated JS ignores all the types, so it ends up passing an array of doubles into `getSurface` which is expecting floats, but as long as that's OK all the way down, I guess that's OK...

src/stage/stage.ts
garyo: sure -- didn't realize how many `@ts-ignore` comments were already in this codebase.

src/component/structure-component.ts
garyo: Hmm, I can't make that work. I tried adding it there, but I still get a tsc err, because (as I understand it, admittedly imperfectly), `defaultAssembly` would be needed in all elements of `StructureRepresentationParametersMap` (i.e. all mapped values of `K`), not just `StructureRepresentationParameters`.
```
% tsc
src/component/structure-component.ts:261:12 - error TS2339: Property 'defaultAssembly' does not exist on type 'Partial<StructureRepresentationParametersMap[K]>'.

261     params.defaultAssembly = this.parameters.defaultAssembly
               ~~~~~~~~~~~~~~~
```

garyo: But this does work (with adding `defaultAssembly` to `StructureRepresentationParameters`):
```typescript
addRepresentation <K extends keyof StructureRepresentationParametersMap>(
    type: K,
    params: Partial<StructureRepresentationParametersMap[K]>|Partial<StructureRepresentationParameters> = {},
    hidden = false
  ) {
    params.defaultAssembly = this.parameters.defaultAssembly
```
Would that be OK?

src/viewer/viewer.ts
garyo: clearTarget is deprecated (as of r97), so I thought it best to remove it. From the [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide#r96--r97):

> WebGLRenderer.clearTarget() is now deprecated. Use WebGLRenderer.setRenderTarget() in combination with WebGLRenderer.clear() instead.

It produces a warning on the console if you use it.

src/viewer/viewer.ts
garyo: Yup, looks like they fixed that. Thanks!

src/viewer/viewer.ts
garyo: It's deprecated and prints a warning. See the [migration guide](https://github.com/mrdoob/three.js/wiki/Migration-Guide#r96--r97).

src/component/structure-component.ts
ppillot: Well, I've checked and it appears that some representations such as label extend their parameters from `RepresentationParameters` and not `StructureRepresentationParameters` which might be the cause.
After better reading at the code it appears that the `defaultAssembly` parameter is not exposed to the user (ie, the user can not set it via `addRepresentation`), so it does not really make sense to add it to the list of parameters in `StructureRepresentationParameters`. 
Your initial typings now seems to me like the best thing to do. Sorry for the noise...

src/surface/volume.ts
ppillot: Yes, that was my guess too. The code was written first with these careful assumptions. As long as we don't meddle with the underlying js logic, it should continue to work Ok.

src/viewer/viewer.ts
ppillot: Ok, cool

src/stage/stage.ts
ppillot: As I said, just a matter of taste here ;-) It makes sense to leave it as is.

src/component/structure-component.ts
garyo: No problem -- always good to try for the best solution.

ppillot:
LGTM!

PR #720, opened 2019-12-31
Fixes for recent three.js r111 update

garyo:
* Fix mouse-wheel zoom direction
* Replace deprecated geometry `.addAttribute` calls (deprecated in three.js r110)

The mouse-wheel zoom fix might be overcomplicated but I did try to handle all cases.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -105,9 +105,9 @@
       "dev": true
     },
     "@types/jest": {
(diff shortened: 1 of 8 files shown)


garyo:
@arose or @ppillot could you take a quick look at this? I'd like to get all recent PRs merged.

PR #724, opened 2020-02-18
In MolecularSurfaceRepresentation, create a double sided surface only when needed

garyo:
This patch tries not to double the amount of geometry created by NGL in surface mode; currently it always creates a `DoubleSidedBuffer` even if the "side" param is not "double". This patch makes it only create the `DoubleSidedBuffer` when requested.

One thing I've noticed is that changing the "side" param does not regenerate the geometry; seems like (at least with this patch) it should. @arose if you could take a look at this I'd appreciate it.

### DIFF
--- a/src/representation/molecularsurface-representation.ts
+++ b/src/representation/molecularsurface-representation.ts
@@ -274,9 +274,13 @@ class MolecularSurfaceRepresentation extends StructureRepresentation {
         })
       )
 
-      const doubleSidedBuffer = new DoubleSidedBuffer(surfaceBuffer)
-
-      bufferList.push(doubleSidedBuffer)
+      if (this.getBufferParams().side == 'double') {
+        const doubleSidedBuffer = new DoubleSidedBuffer(surfaceBuffer)
+        bufferList.push(doubleSidedBuffer)
+      }
+      else {
+        bufferList.push(surfaceBuffer)
+      }
     }
 
     return { bufferList, info } as StructureRepresentationData
--- a/src/stage/mouse-observer.ts
+++ b/src/stage/mouse-observer.ts
@@ -227,23 +227,34 @@ class MouseObserver {
    * @param  {Event} event - mouse event
    * @return {undefined}
    */
-  _onMousewheel (event: WheelEvent) {
+  _onMousewheel(event: WheelEvent & { wheelDelta?: number, wheelDeltaY?: number }) {
     if (event.target !== this.domElement || !this.handleScroll) {
       return
     }
     event.preventDefault()
     this._setKeys(event)
 
-    var delta = 0
-    if (event.deltaY) {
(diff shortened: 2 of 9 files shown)


garyo:
@arose could you take a look at this?

src/representation/molecularsurface-representation.ts
```diff
@@ -274,9 +274,13 @@ class MolecularSurfaceRepresentation extends StructureRepresentation {
...
-      const doubleSidedBuffer = new DoubleSidedBuffer(surfaceBuffer)
-
-      bufferList.push(doubleSidedBuffer)
+      if (this.getBufferParams().side == 'double') {
+        const doubleSidedBuffer = new DoubleSidedBuffer(surfaceBuffer)
+        bufferList.push(doubleSidedBuffer)
+      }
+      else {
+        bufferList.push(surfaceBuffer)
+      }
```
fredludlow, line 283: @garyo - this looks great, but one thing I don't quite understand here. Stick a couple of breakpoints at line 278 and 282 to see what I mean:

If I load the representation/molecularsurface-av example it draws two surfaces in contour mode (so this section of code isn't reached at all - so far so good). 

Then if I set the side to 'front' and turn off contour (in that order), it just pushes "surfaceBuffer" onto bufferList (no DoublesidedBuffer - still good).

Then if I switch the side to 'double' - it just works! but DoubleSidedBuffer was never called... So, is DoubleSidedBuffer actually not required?

NB: If I load in contour mode, leave it on double and turn off contours, a DoubleSidedBuffer is created, but I can't tell the difference visually between the two results..

src/stage/mouse-observer.ts
```diff
@@ -227,23 +227,34 @@ class MouseObserver {
...
+      // Old Firefox or IE 11: deltaY but no deltaMode; treat as pixels
+      delta = -event.deltaY * (2.5 / 100.0)
+    } else if (event.wheelDelta !== undefined) {
+      delta = -event.wheelDelta * (2.5 / 100)
+    } else if (event.wheelDeltaY !== undefined) {
+      delta = -event.wheelDeltaY * (2.5 / 100)
     } else {
-      // Firefox or IE 11
-      delta = -event.deltaY / (event.deltaMode ? 0.33 : 30)
+      // Old Firefox, MouseWheelEvent
```
fredludlow, line 256: Nice :)

arose:
> This is working fine for me, but see my comment in molecularsurface-representation.ts. I haven't dug that far but I'm wondering what DoubleSidedBuffer is required for - presumably it does some things that aren't done by using FrontSide, BackSide, DoubleSide from three?

The idea with DoubleSidedBuffer is avoid some artifacts when rendering transparent surfaces and is used instead of DoubleSide. It renders the geometry twice once with FrontSide and once with BackSide.

Issue #617, opened 2019-01-24
Rebuild of documentation


harryjubb:
The build for the[ automated documentation](http://nglviewer.org/ngl/api/identifiers.html) is for quite an old version now (1.0.0-beta7). Are there any plans to update it to cover the latest release? If it's PR-able, happy to help.

@fredludlow @jamesfoley

fredludlow:
`npm run doc` currently fails for me with typescript errors. I can have a look at those and try and fix it. @arose - assuming we get the doc command working, how easy is it to update the website?

arose:
Last time I tried typedoc would not handle the typescript version NGL was written in. Maybe things are better now. I can update the website if the doc cmd is working again.

bergwerf:
I was also unable to build the documentation. This is the output I got: https://pastebin.com/Dsxb7Btx
All errors in THREE and ngl.

arose:
`typdoc` needs to be updated, the version used in NGL is not compatible with the TS version used in NGL.

bergwerf:
I can confirm that updating `typedoc` to the latest version and adding `allowSyntheticDefaultImports` (to import `@types/promise-polyfill/index`) to the tsconfig indeed fixes documentation generation. I can create a PR with just this but I also saw a few other dependencies need to be updated (some deprecation warnings and things moving to a different package). Is it helpful if I create a PR with some `npm update` and dependency changing, or might that break things?

fredludlow:
@bergwerf - I'd be very happy with either option, if you can get deps updated cleanly then that'd be brilliant, if not then just getting the docs done would be very helpful.

bergwerf:
There is also https://github.com/arose/ngl/pull/726 but that doesn't cover typedoc.

PR #726, opened 2020-02-19
Fix/update package dependencies

garyo:
This removes some dependencies on vulnerable old packages such as old versions of core-js.
This should apply cleanly on top of my previous pull request -- only the last commit, ece8274a, is part of this one. If not, I'll rebase and push again.


bergwerf:
@ppillot could you look at this? I would like to update more dependencies (in particular typedoc, see #617) but perhaps it is easier if this is merged first.

fredludlow:
There's a minor regression from #716, which introduces `WebGL2RenderingContext`. This isn't supported in IE11 or Safari on iOS, but the only times it is used are to access the DYNAMIC_DRAW enum member, which is also available on WebGLRenderingContext. So, rather than make a whole PR and cause rebasing faff, do you think you could add the change (there's 6 occurences of WebGL2RenderingContext->WebGLRenderingContext in buffer.ts)?

garyo:
> do you think you could add the change (there's 6 occurences of WebGL2RenderingContext->WebGLRenderingContext in buffer.ts)?

Sure, no problem. Will do.

fredludlow:
Thanks! Do you want to hit merge (I guess on the others first?)

PR #739, opened 2020-04-16
Fix/clipping in camera clip mode

garyo:
More clipping handling; see PR #656

This code block didn't get updated as part of the previous clipping updates.
Camera near/far should not get reset in camera clip mode, just obey params directly.


garyo:
Sorry for the stacked PRs, hope that's OK. That's all for now anyway!

PR #740, opened 2020-04-21
Add a new representation 'dihedral-histogram' and an example of usage

rudolfspetrovs:
This is a new representation.

It creates a circular histogram that shows distribution of dihedral angles.
You need to provide actual histogram data and the four atoms of the torsional angle.


fredludlow:
Full disclosure: Rudolfs is at Astex with me, we've been using this internally and find it useful. If no-one objects I'll merge it but I'll leave it open for a week or so for feedback and/or objections - I'm conscious of not wanting to railroad any changes!

arose:
Sounds interesting, who does it look in actin?

rudolfspetrovs:
Well, I/we imagined this representation for ligands, did not think of using it for proteins.
You need to provide data of torsional angle distribution, for example from some source, like CCDC DB (Mogul), or you could QM calculations of torsion to get the according distribution.
However, Fred suggested that you could use Ramachandran plot data and visualise that information on the protein. But we have not tried it yet.

arose:
Sorry, I meant 'look in action' :)

rudolfspetrovs:
Here is a screenshot of it.

image004

arose:
Thank you @rudolfspetrovs , that looks very useful

PR #781, opened 2020-08-14
Restore test suite

fredludlow:
I've converted the old test suite to use jest and ts-jest and they're all passing (some were previously being skipped, this is still the case).

There are also a few changes to the source code (e.g. I've made parser parameters optional and changed some types from specific array types to more general NumberArray types)


garyo (approved):
This is excellent! Tests all pass for me, on this branch and when merged into my local latest (linear color workflow etc.).
This will be really helpful going foward; thanks!

PR #785, opened 2020-09-13
Fix regexp for SDF extra data

gky360:
This PR fixes the regexp used for parsing extra data in SDF files.

According to [SDF spec showed in Wikipedia](https://en.wikipedia.org/wiki/Chemical_table_file#SDF),  extra data items in SDF files are denoted as follows.
The number of spaces between `>` and `<` is 2.

```
>  <Unique_ID>
XCA3464366
 
>  <ClogP>
5.825

>  <Vendor>
Sigma

>  <Molecular Weight>
499.611
```

However, [the current regexp that is used to search for metadata (`/> <(.+)>/`)](https://github.com/nglviewer/ngl/blob/v2.0.0-dev.37/src/parser/sdf-parser.ts#L11) contains only 1 space between `>` and `<`.

Therefore, I fixed the regexp to match the above example too.

### DIFF
--- a/src/parser/sdf-parser.ts
+++ b/src/parser/sdf-parser.ts
@@ -8,7 +8,7 @@ import { Debug, Log, ParserRegistry } from '../globals'
 import { assignResidueTypeBonds } from '../structure/structure-utils'
 import StructureParser from './structure-parser'
 
-const reItem = /> <(.+)>/
+const reItem = /> +<(.+)>/
 
 class SdfParser extends StructureParser {
   get type () { return 'sdf' }


fredludlow:
Thanks for this @gky360 - the spec is actually pretty horrible (see below) but this is an improvement on our current implementation!

See
https://depth-first.com/articles/2020/07/13/the-sdfile-format/
and
http://help.accelrysonline.com/ulm/onelab/1.0/content/ulm_pdfs/direct/reference/ctfileformats2016.pdf

All these are valid (as well as the double-spaced version above).
> <Melting_Point>
><IC_50>(JNJ-5207852)
> DT13
>Arbitrary Data Goes Here<pKi_HISTAMINE_H3>

fredludlow (approved):

PR #792, opened 2020-09-23
693 negative iso transparency

### LINKED ISSUE #693: surface repr: opacity does not work with neglateIsolevel
Hi, please try with the .cube file in this https://github.com/arose/nglview/files/3579869/input.zip
- load it to http://nglviewer.org/ngl/
- With the surface repr, click "neglateIsolevel" and then adjust the "Opacity". There is no effect.

Here is the original context: https://github.com/arose/nglview/issues/785#issuecomment-528413612

thanks and cheers

fredludlow:
I think this finally fixes #693. Triangle orders and normal vertices are flipped when contouring at a negative level.

Demo here:
https://fredludlow.com/ngl/fix-693/examples/webapp.html?script=test/negative-iso

Built file available here:
https://fredludlow.com/ngl/fix-693/build/js/ngl.dev.js


src/surface/marching-cubes.ts
```diff
@@ -744,8 +751,8 @@ function MarchingCubes (this: MarchingCubes, field: number[], nx: number, ny: nu
         }
       } else {
         // FIXME normals flipping (see above) and vertex order reversal
```
garyo, line 753: This looks great, thanks Fred! I think you could delete this FIXME comment with this fix.

garyo (approved):
LGTM!

src/surface/marching-cubes.ts
fredludlow: Done! Thanks

Issue #788, opened 2020-09-16
Coordinate systems inconsistencies


hoba87:
I have observed some inconsistencies/bugs regarding the coordinate system, when combining a structure loaded from file and a shape. Three subissues I have observed:
* the viewed position of the shape object / shapeComp representation, already for (0,0,0), depends on the rotation angle (changing it simultaneously). 
* it seems that the minimum of the given shape coordinates are mapped to (0,0,0) or something different of the structure loaded from the file
* there is some issue with at least the z axis direction, it seems that the direction of the z-axis is opposed to the structure loaded from file

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
</head>
<body style="background-color:grey; margin:0; padding:0; overflow:hidden">
  <script src="ngl.js"></script>
  <script>
    document.addEventListener("DOMContentLoaded", function () {
      var stage = new NGL.Stage("viewport");
      stage.loadFile("ngl.pdb").then(function(o) {
        o.addRepresentation("spacefill", {
          side: "front", sphereDetail: 0, radiusType: "covalent", scale: 0.5, colorScheme: "element",
        });
        o.autoView();
        o.setScale(1);
        o.setRotation([-1.5707963267948966,0,-1.5707963267948966]);
        o.setPosition([0,0,0]);
      });

      var shape = new NGL.Shape("shape");
      shape.addMesh(
        [ 1, 1, 1, 1, 2, 1, 1, 1, 2, 1, 2, 2, 1, 2, 1, 1, 1, 2 ],
        [ 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0 ],
        undefined, undefined, "surface"
      );
      var shapeComp = stage.addComponentFromObject(shape);
      shapeComp.addRepresentation("buffer");
      shapeComp.setScale(1);
      shapeComp.setRotation([-1.5707963267948966,0,-1.5707963267948966]);
      shapeComp.setPosition([0,0,0]);

      stage.setParameters({cameraType: "orthographic"});
    });
  </script>
  <div id="viewport" style="width:800px; height:600px;">
  </div>
</body>
</html>

```

fredludlow:
I'm finding this a bit hard to reproduce (though you may well be right). Can you provide an example with the behaviour you'd expect and what you actually see?

[...]

PR #799, opened 2020-09-28
hint for setTransform for global rotation in setRotation doc

hoba87:
Should avoid possible new issues like #788


fredludlow (approved):
LGTM - thanks @hoba87!

PR #807, opened 2020-11-10
Avoid divide-by-zero when normalizing vectors

garyo:
There's no "right" answer for what to do when normalizing a (0,0,0)
vector, but NaNs is definitely not a good result. This just leaves
the vector as (0,0,0), non-normalized but numerically valid.

Zero-length vectors can result from creating normals when building
isosurfaces, when a triangle is zero area.

I ran across this with a SARS-Cov2 model, making a vws surface. I think
it creates a zero-area triangle which triggers these NaNs in the triangle 
normals and positions. You don't notice them normally when just
viewing the results (three.js seems to be fairly forgiving, but I export
glTFs and they are fatal there.

### DIFF
--- a/src/math/vector-utils.ts
+++ b/src/math/vector-utils.ts
@@ -146,10 +146,14 @@ export function normalizeVector3array (a: Float32Array) {
     const x = a[ i ]
     const y = a[ i + 1 ]
     const z = a[ i + 2 ]
-    const s = 1 / Math.sqrt(x * x + y * y + z * z)
-    a[ i ] = x * s
-    a[ i + 1 ] = y * s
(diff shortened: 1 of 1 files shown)


fredludlow (approved):
This looks good, and is consistent with how three.js handles it: https://github.com/mrdoob/three.js/blob/dev/src/math/Vector3.js#L471

I wonder if we should update `v3normalize` as well for consistent behaviour?

I wonder if we should also update `v3normalize` as well?

garyo:
Thanks -- merged. I'll do a separate PR for `v3normalize`, you're right -- that needs the same treatment.

Issue #775, opened 2020-08-05
Color management in NGL


garyo:
Recent versions of three.js (since r112) have enabled a proper linear color workflow. I'm working on implementing that; I have most of the core color management working using the recommendations in Don McCurdy's [Color Management in three.js](https://www.donmccurdy.com/2020/06/17/color-management-in-threejs/) doc. 

My question for other users is, should I just turn it on always? The results should be quite similar, though definitely not identical. Colors overall should be more accurate, especially where they are blended (in surface representations, for example). It's simpler to implement without having to maintain the old path.

I can enable it via a switch in the viewer, but passing that state around everywhere is nontrivial -- if users would be satisfied with just switching over, that would be perfect for me. Opinions or thoughts?

fredludlow:
It's not something I have any expertise in, but if it's an improvement and doesn't break things then it sounds good to me! Do you have a demo somewhere?

garyo:
This took me ages to get to, but here's a demo of color-managed NGL viewer: http://oberbrunner.com/ngl/examples/webapp.html
The main difference you should see is that the current method darkens the darks excessively, and adding proper color management keeps them from being quite so crushed. You can see it clearly if you load a molecule like 2c8n and turn the metallic almost all the way up in the ball+stick ligand view. With the existing version, it's pretty much black except for the brightest highlights. With the color-managed view you can see the sticks and the balls.
As another test, try 1d3e (I like to set vws surface and scale factor of 0.7); the existing version is super dark around the edges of the atoms. Some people will like the excessive contrast of the current version, to be sure, and by adjusting the gamma you can get back to that look. 
In any case my color management code can be enabled/disabled in the API.

Opacity is where you see it most clearly. Here's 4rik with ball+stick and a 50% opacity default surface, without color management:
![image](https://user-images.githubusercontent.com/139975/92950822-58def800-f42b-11ea-9289-bdfab614546d.png)

and the same, with color management:
![image](https://user-images.githubusercontent.com/139975/92950838-61cfc980-f42b-11ea-8c20-5ef6d4346254.png)

Here's 1d3e with a surface (vws, scale=0.7) with no color management:
![image](https://user-images.githubusercontent.com/139975/92949985-f6d1c300-f429-11ea-9b74-26a60c926779.png)

and with color management:
![image](https://user-images.githubusercontent.com/139975/92949946-e7eb1080-f429-11ea-807e-b33eaaf95351.png)

[...]

PR #808, opened 2020-11-10
Color management: linear workflow

garyo:
Here's my color-management code -- linear workflow internally, turned on by default.

I've been using it for a while and it works well for me. There are two issues to discuss, I think:
1. How to allow users to specify the color encoding, if they want to go back to the default? I've added a `colorEncoding` param to the viewer, which seems like a good place for it, but I don't have a good idea on how to thread it through to colormaker.ts where it's needed. So right now that param does nothing -- in order to get back to the old workflow you have to change `colorSpace` in colormaker.ts. Ideas appreciated.

2. @arose commented in #775 that it might be better to implement color management in the shader; I'd like to propose that it's at least OK if not better to do it this way because the values (vertex colors, param values, and textures) are explicitly linear, which works properly e.g. when exporting models as glTF. The only real issue is that in colormaker.ts, it's still creating 8-bit linear color values because it uses `.toHex`. Ideally it should produce float rgb triples and use those everywhere, but that's a bigger change than I wanted to get into for this PR. It does mean you lose some precision when specifying very dark colors, but I don't think that's so bad for now; moving to full-float colors would be a good next step.

### DIFF
--- a/src/color/atomindex-colormaker.ts
+++ b/src/color/atomindex-colormaker.ts
@@ -6,7 +6,7 @@
 
 import { ColormakerRegistry } from '../globals'
 import { defaults } from '../utils'
-import Colormaker, { StuctureColormakerParams, ColormakerScale } from './colormaker'
+import Colormaker, { StuctureColormakerParams, ColormakerScale, manageColor } from './colormaker'
 import AtomProxy from '../proxy/atom-proxy'
 import ModelProxy from '../proxy/model-proxy'
 
@@ -47,6 +47,7 @@ class AtomindexColormaker extends Colormaker {
    * @param  {AtomProxy} atom - atom to get color for
    * @return {Integer} hex atom color
    */
+  @manageColor
   atomColor (atom: AtomProxy) {
     return this.scalePerModel[ atom.modelIndex ](atom.index)
   }
--- a/src/color/bfactor-colormaker.ts
+++ b/src/color/bfactor-colormaker.ts
@@ -5,7 +5,7 @@
  */
 
 import { ColormakerRegistry } from '../globals'
-import Colormaker, { StuctureColormakerParams, ColormakerScale } from './colormaker'
+import Colormaker, { StuctureColormakerParams, ColormakerScale, manageColor } from './colormaker'
 import AtomProxy from '../proxy/atom-proxy'
 import Selection from '../selection/selection'
 
@@ -52,6 +52,7 @@ class BfactorColormaker extends Colormaker {
     this.bfactorScale = this.getScale()
   }
 
+  @manageColor
   atomColor (a: AtomProxy) {
     return this.bfactorScale(a.bfactor)
   }
--- a/src/color/chainid-colormaker.ts
+++ b/src/color/chainid-colormaker.ts
@@ -5,7 +5,7 @@
  */
 
 import { ColormakerRegistry } from '../globals'
-import Colormaker, { StuctureColormakerParams, ColormakerScale } from './colormaker'
+import Colormaker, { StuctureColormakerParams, ColormakerScale, manageColor } from './colormaker'
 import AtomProxy from '../proxy/atom-proxy'
 import ChainProxy from '../proxy/chain-proxy'
 import ModelProxy from '../proxy/model-proxy'
@@ -41,6 +41,7 @@ class ChainidColormaker extends Colormaker {
     })
   }
 
+  @manageColor
   atomColor (a: AtomProxy) {
     const chainidDict = this.chainidDictPerModel[ a.modelIndex ]
     return this.scalePerModel[ a.modelIndex ](chainidDict[ a.chainid ])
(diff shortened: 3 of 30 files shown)


garyo:
@fredludlow any thoughts on this? I'd like to get it merged at some point.

fredludlow:
Sorry - I made a few false starts at looking at it earlier.

The transparent surface example you put in #775 looks great, 

I'm slightly concerned that we lose a bit of the 'depth' effect, e.g. here (managed):
image

vs original:
image

I find it harder to perceive 3D structure from a static image in the linear color managed version - I guess because the darkening of the curved surfaces is less extreme they don't seem to pop out of the image as much - is this by design, or something to do with the loss of precision? (Or might it be fixed/compensated for by updating the lighting model - or maybe it's just what I'm used to and I'd adapt if I tried it for a few weeks...?)

fredludlow:
ps - Built versions here (this PR: https://fredludlow.com/ngl/pr-808/examples/webapp.html?script=color/atomindex ) and current master https://fredludlow.com/ngl/master/examples/webapp.html?script=color/atomindex ) in case anyone wants to see side-by-side. If you stick an 'av' surface at ~70% opacity the linear one looks significantly nicer!

garyo:
Hi Fred; thanks so much for putting up these builds for comparison! You're right that in the lighter colors there's somewhat less contrast and there's less over-darkening around the edges. It's more physically accurate but you're also right that sometimes inaccurate can look better... 

Currently ambient light defaults to 0.2; I think that's partly to compensate for the fact that the mids and darks were too dark otherwise. Perhaps along with this PR I should default ambient lower, like 0.05 (or even off altogether). That gets back some of the contrast. Also with a linear workflow that makes more phyical sense. What do you think?

Adding fog can help too (fog should be more accurate now). Try lowering the Fog Near param. Also adding a bit of metalness always looks good; that should bring back some of the darkening around the edges due to the Fresnel effect. But I don't know that we want to change those defaults; they'll be kind of specific to the model.

fredludlow:
I've asked a few of my users, and one came up with a view sort of like this (similar to the images above but more typical of what they'd look at day-to-day):

**Linear Color** (with ambientIntensity: 0)
https://fredludlow.com/ngl/pr-808/examples/webapp.html?script=test/binding-site
image

**Original Version**
https://fredludlow.com/ngl/master/examples/webapp.html?script=test/binding-site
image

Their view was the linear color version looks 'flatter' and a bit more cartoony, which I think is what I was trying to get across above. This is more of an issue if you're preparing images for publication (when you're working 'live' you adjust the rotation constantly to help maintain your 3D mental model). 

In your original post you proposed having the option to switch between modes, which would be my preference if it's doable. You said you weren't sure how to thread the value through - I'm not sure I have any great ideas, but is it something you'd want to allow different settings on different representations/buffers or is it a global setting? (I'm guessing global as you change renderer.outputEncoding?)

garyo:
Thanks Fred -- user feedback is invaluable for this. I'll work on creating an option. Yes, it has to be global -- it affects the colormakers as well.

fredludlow:
Apologies once again for the big time-lag... I'm looking at this now and wondering if you can help me understand it a bit...

In the master branch, `renderer.outputEncoding` etc are set to 3000 (which is `LinearEncoding`). The `manageColor` decorator doesn't exist here so it's sort of equivalent to having called `setColorEncoding(sRGBEncoding)`

With a fresh build from your branch the outputEncoding is set to 3001 - `sRGBEncoding` and `colorSpace` (`setColorEncoding`) is `linear` - the overall brightness is correct but I'm getting the slightly flat look described above,

I can recreate the original behaviour with `stage.setOutputEncoding(3000 /* Linear */)` and `stage.setColorEncoding(3001 /* sRGBEncoding */)` - this makes sense I think...


So, to summarize:

The old way is working in sRGB internally with no transformation upon output, the properly managed way is to work with linear colors internally and transform to sRGB on output? 

*If* I'm right (big if :) ), is there any reason why you'd ever want to `setOutputEncoding` and `setColorEncoding` to the same value? If not, should there be a single parameter which switches between the two schemes?

If I've got this all wrong please let me know! (Happy to have a call if it's easier than my long-delayed responses..)

garyo:
Thanks for the review, @fredludlow . Yes, you're exactly right in your summary. My intent with the last commit 3652a347 was to set the default behavior back to the traditional way (sRGB colors internally, compositing in sRGB, no output conversion), but looks like perhaps I didn't get it all the way done. 
And yes, you have a good point about just using a single 'setColorWorkflow' method in viewer.ts. I didn't go that way originally because the output setting was in the viewer and the colormaker setting was... complicated. But now that they're both in the viewer, it makes sense to collapse them.
I'll fix both of those issues.

garyo:
Fixed defaults, and tested default behavior vs. master and it looks the same. 
Now there's only one new API function, `viewer.setColorWorkflow("linear"|"sRGB")`.

fredludlow (approved):

fredludlow:
Thank you for the fix @garyo, now merged!

PR #811, opened 2020-11-19
added polarh and apolarh selection keywords

rjdkmr:
This adds two new selection keywords "polarh" and "apolarh" to display polar (-X-H) and non-polar (-C-H) hydrogen atoms.

### DIFF
--- a/src/selection/selection-constants.ts
+++ b/src/selection/selection-constants.ts
@@ -25,14 +25,15 @@ export enum kwd {
   RING = 17,
   AROMATICRING = 18,
   METAL = 19,
-  NONE = 20
+  NONE = 20,
+  POLARH = 21
 }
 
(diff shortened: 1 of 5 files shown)


src/selection/selection-constants.ts
```diff
@@ -25,14 +25,15 @@ export enum kwd {
   RING = 17,
   AROMATICRING = 18,
   METAL = 19,
-  NONE = 20
+  NONE = 20,
```
garyo, line 28: I think NONE should always be last here.

rjdkmr: I have fixed it in latest commit. Thanks.

fredludlow: Thanks @rjdkmr - all looks good I think.

fredludlow (approved):

PR #813, opened 2020-11-28
Position tooltip relative to containing element

yitzchak:
Current positioning of tooltip appears to assume that the stage is in fullscreen mode. Updated to use bounding rectangle of containing element.


garyo (changes requested):
This seems like a good idea, but when I tried this on the `annotations.html` example, it looks like it draws the tooltips in the wrong place with your patch. To test it, run a webserver (e.g. `python -m http.server`) and go to `http://localhost:8000/examples/annotations.html` and look at the tooltips.

yitzchak:
I see that. For some reason the patch works when using ngl inside Jupyter. I'm going to close this and if I manage to figure out why this is and if a patch is even needed I open a new PR.

PR #823, opened 2021-01-08
added a machine learning visualization tool to examples

bralexander:
I've added a machine learning visualization tool, and a residue clicking functionality that toggles ball and stick representation of a selection. 
I could add some more machine learning csv's so you can view more than one protein if you're interested.
Shout out to Fred for answering my questions!


examples/scripts/interactive/compvis-viewer.js
```diff
@@ -0,0 +1,603 @@
...
+    NGL.autoLoad(csvFile, {
+      ext: 'csv',
+      delimiter: ' ',
+      comment: '#',
+      columnNames: true
+    })
+  ]).then(function (ol) {
+    struc = ol[0]
+    csv = ol[1].data
+
```
fredludlow, line 173: Rather than iterating through csv in the `atomColor` function below, might it be better to build a lookup object here

```javascript
var residueData = {}
for (var i = 0; i < csv.length; i++) {
    resNum = parseFloat(csv[i][csvResNumCol]
    residueData[resNum] = csv[i]
}
```

examples/scripts/interactive/compvis-viewer.js
```diff
@@ -0,0 +1,603 @@
...
+    firstResNum = parseInt(csv[0][csvResNumCol])
+
+    heatMap = NGL.ColormakerRegistry.addScheme(function (params) {
+      this.atomColor = function (atom) {
+        for (var i = 0; i < csv.length; i++) {
```
fredludlow, line 178: Then here: 
```javascript
const csvRow = residueData[atom.resno]
const wtProb = parseFloat(csvRow[csvWtProbCol])

// Etc...
```

fredludlow:
Hi @bralexander

Thanks for this, I like the example! 

I've got a couple of comments above about removing a loop within `atomColor` method.

I also think you could possibly simplify quite a lot of your color generation code by following the example in showcase/qmean specifically around line 41-50: 
https://github.com/nglviewer/ngl/blob/963b4fecde843fdf80c580e8f7bd55989db6a6f6/examples/scripts/showcase/qmean.js#L41-L50

There was a bug in CSV parsing which had meant that the qmean example didn't work (it used a space delimited file), but actually in your case the parsing is fine so you should (fingers crossed) be able to get it to work that way.

examples/scripts/interactive/compvis-viewer.js
bralexander: I am able to get all of these suggestions working on some proteins, but not on 2isk and vice-versa. the problem is that the csv data does not always match up with the structure data. For instance I was using atom.residueIndex as the iterator ( instead of atom.resno -- I changed to residueData accordingly), but atom.residueIndex is 219 long while the csv is 218 long. I tried your way, with resNum and resno, and had same issue on a different protein.

If the csv and structure data do not match does this require me to use csv.length as the limit to iteration (i.e. for loop in atomColor) to avoid getting "Undefined is not an object" error? Or is there another way that I'm missing?

I can use qmean example to simplify color generation, but with a for loop.

I'll try to have an updated example where you can switch proteins easily pushed by tomorrow

examples/scripts/interactive/compvis-viewer.js
fredludlow: Looking at the other PDB you pushed (6ar1) you've got data for 2 chains (so the residue numbers restart). I'd update this to use the combination of chain name and residue number to index `residueData`, e.g.:

```javascript
var residueData = {}   // Ends up as { 'A101': csv[x], 'A102': csv[y], ...}
for (var i = 0; i < csv.length; i++) {
    chain = csv[i][chainCol] // chainCol is 3 in your example, `chain` will be something like 'A'
    resNum = parseFloat(csv[i][csvResNumCol] // e.g. 101
    residueData[chain+resNum] = csv[i]  // chain+resNum will be 'A101'
}
```

examples/scripts/interactive/compvis-viewer.js
```diff
@@ -0,0 +1,640 @@
...
+      this.domain = [ 0.5, 1 ]
+      this.scale = 'rwb'
+      this.mode = 'rgb'
+      var scale = this.getScale()
+      this.atomColor = function (atom) {
+        //console.log(atom.residueIndex)
+        // var index = parseFloat(residueData[atom.residueIndex][0])
+        // console.log(atom.residueIndex, index)
+        const csvRow = residueData[atom.resno]
+        const wtProb = parseFloat(csvRow[csvWtProbCol])
```
fredludlow, line 195: I'm guessing this is the line that caused the error? Your CSV file doesn't contain data for every residue in the PDB file, so I *think* you just need to decide what the appropriate colour to return is when there's no data, e.g.

```
const csvRow = residueData[atom.resno] // Or `residueData[atom.chainname + atom.resno]
if (csvRow === undefined) {
  return 0xcccccc
}
// Then as you have it there
const wtProb = parseFloat(csvRow[csvWtProbCol]) 
...
```

examples/scripts/interactive/compvis-viewer.js
fredludlow: See my comment in your updated version - I think you just need to check cases where residueData[atom.resno] is undefined and return a default color.

examples/scripts/interactive/compvis-viewer.js
bralexander: Got it working thanks! 
I am going to clean up the code, and add some helpful features and bug fixes. I should be done in a few hours.

bralexander:
I have pushed the best working version with your suggestions, some bug fixes, a few helpful features, and one spare machine learning csv. 
I have built it locally and standard  --fixed things, but for some reason it won't let me push anymore. I'll have to figure this out tomorrow.

bralexander:
Everything done ready to go!

fredludlow (changes requested):
Thanks again @bralexander - this really looks great. I've got some very minor comments:

1) Out of interest - how was the CSV data generated and what is it showing? And just to be paranoid, are there any restrictions on how the results are allowed to be shared/distributed?

2) You can remove the default click listener (Animation currently ends up firing twice when you click on an atom for me). Putting a call to `stage.mouseControls.remove('clickPick-left')` before you add your handler should do it. 
 
3) Can you run npx standard --fix examples/scripts/interactive/compvis-viewer.js please :)

4) Feel free to stick an author line in the example files

Thanks again!

ps - I'll squash the commits into 1 when I merge unless you want me to preserve them all? (a few PDB files were added and removed from your branch I think which adds a bit of weight to the repository).

bralexander:
- Question for you -- I had been coloring the pocket surface repr by heatMap color but that no longer works now that I am using the 'rwb' scale. I think this occurs because the 'rwb'  scale does not return hex color values anymore. CustomPercent still works which assigns color values. Is there any way to fix this that you can think of? Custom percent color is fine for the pocket color if not. 

Per your questions:
- Here's two papers about the machine learning net https://pubs.acs.org/doi/abs/10.1021/acssynbio.0c00345 & https://www.biorxiv.org/content/10.1101/833905v1.abstract. 

- The neural net is available for use by educational institutions and those associated at mutcompute.com. I will be adding this tool to the website soon.

- Essentially how it works is it looks at a bunch of proteins and then from that knowledge gives you a probability for the WT (wild type) amino acid, and then predicts an amino acid that should be there and assigns it a probability. 

- The two csv's (2isk and 2gv5) that I pushed I have permission to use and share, but the others (6ar1 etc) I do not have permission to share. I'm pretty sure I removed those already.

- And yes please squash the commits.

fredludlow:
> Question for you -- I had been coloring the pocket surface repr by heatMap color but that no longer works now that I am using the 'rwb' scale. I think this occurs because the 'rwb' scale does not return hex color values anymore. CustomPercent still works which assigns color values. Is there any way to fix this that you can think of? Custom percent color is fine for the pocket color if not.

I wasn't sure either, but after a bit of digging, it turns out that the qmean example is also not quite doing what I'd thought. Specifying `this.domain`, `this.scale` and `this.mode` in the colormaker constructor doesn't actually do anything, and you might not want to hard code these anyway. They will get passed through from the representations `colorXYZ` parameters. (There's a bit of manipulation going on in `ColormakerRegistry` [here](https://github.com/nglviewer/ngl/blob/master/src/color/colormaker-registry.ts#L204-L213) that ultimately sets all this up.

To fix your example, you have two options:

1) If you want to control the parameters completely (and not allow them to be overridden by the representation's `colorScale`, `colorDomain` and `colorMode` params) you can write something like:

```javascript
NGL.ColormakerRegistry.addScheme(function (params) {
    this.parameters = Object.assign(this.parameters, {     // Using Object.assign preserves other attributes that might be important
      domain: [ 0.5, 1 ],
      scale: 'PuBu',
      mode: 'rgb'
    }) 
    var scale = this.getScale()
    this.atomColor = function(atom) {...
```
(I think the qmean example should be updated like this with a comment as to what it's doing, as well as a reference to it in http://nglviewer.org/ngl/api/manual/coloring.html)

2) If you want scale, domain etc to follow representation parameters, you can remove the parameter settings from the colorMaker and make sure that when you create `pocketRepr` it has a `colorScale` parameter supplied, i.e

```javascript
NGL.ColormakerRegistry.addScheme(function (params) {
    // Remove domain, scale, mode from here
    var scale = this.getScale() // Still need this
    this.atomColor = function(atom) {...
```


Thanks also for the info on the method - sounds really interesting.

Apologies this has taken a while - do you want to fix the heatMap thing and I'll merge it all in one go?

bralexander:
Thanks, that fixed the problem. There are a couple minor bugs that I would still like to fix (have to double click residue when custom box is checked, and would be nice to change pocket color when scheme color changes), but I think it's good enough for now. 
No worries on the delay I imagine you are much busier than I. Maybe I'll get around to fixing those bugs before you merge, but if not I think it's good enough.

fredludlow (approved):

Issue #789, opened 2020-09-17
limited atom selection


hoba87:
When I want to select atoms from larger structures (~100.000 atoms) the selection is very limited.
E. g. with following representation with selection, the atoms 0, 1 and 2 are shown, while atoms 1000 and/or 10000 are not shown (the exact limit seems to depend on the number of atoms in the full structure)
```
    o.addRepresentation("spacefill", {
      side: "front", sphereDetail: 0, radiusType: "covalent", scale: 0.5, colorScheme: "element",
      sele: "@0,1,2,1000,10000"
    });
```

fredludlow:
The @ rule uses NGL's internal atomindex, not serial numbers - (though usually that's only off-by-one if your serials start at 1 and don't have gaps).

Have you got an example file you can share?

hoba87:
For the following example files I see only three atoms (1000, 10000, 48000) instead of the selected 6 atoms:

```
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
</head>
<body style="background-color:grey; margin:0; padding:0; overflow:hidden">
  <script src="ngl.js"></script>
  <script>
    document.addEventListener("DOMContentLoaded", function () {
      var stage = new NGL.Stage("viewport");
      stage.loadFile("ngl.pdb").then(function(o) {
        o.addRepresentation("spacefill", {
          side: "front", sphereDetail: 0, radiusType: "covalent", scale: 0.5, colorScheme: "element",
          sele: "@1000,10000,48000,49000,50000,80000"
        });
        o.autoView();
      });
      stage.setParameters({cameraType: "orthographic"});
    });
  </script>
  <div id="viewport" style="width:800px; height:600px;">
  </div>
</body>
</html>
```
The zipped pdb file:
[ngl.pdb.gz](https://github.com/nglviewer/ngl/files/5245432/ngl.pdb.gz)

fredludlow:
Two problems here: 

1) You're hitting the limits of the PDB format again. That file has 180000 atoms in it and some of them have repeated serial numbers. The PDB format only has 5 columns to declare the (unique!) serial number for an atom so can only cope with 99999 atoms. There are hacks to get around this, I think NGL supports hexidecimal atom IDs, see #241, but I've never used this.

[...]

PR #825, opened 2021-01-08
change data type of atomCount to uint32, making selections work for s…

hoba87:
…tructures with more than 2^16 atoms.
Solution for issue #789


fredludlow (approved):
Thanks @hoba87 LGTM

PR #830, opened 2021-01-20
Ensure CSV parameters are not dropped when creating the CsvParser

fredludlow:
The showcase/qmean example was broken (it uses a 'csv' file with space delimiter, and the relevant `delimiter` parameter was not making it all the way to the CsvParser constructor). This is now fixed. The way this is done isn't totally satisfactory (we're telling ParserLoader the names of all possible parser parameters - would make sense to just tell it those that don't need to get passed through), I'll open an issue to make this more sensible, but right now this fixes that example.

Relates tangentially to #823, I'll comment there

### DIFF
--- a/src/loader/parser-loader.ts
+++ b/src/loader/parser-loader.ts
@@ -15,6 +15,9 @@ export interface ParserParams {
   cAlphaOnly?: boolean
   name?: string
   path?: string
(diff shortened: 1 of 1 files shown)


garyo (approved):
LGTM

PR #833, opened 2021-01-20
Correctly parse atomname of form '0C21', fixes #797

### LINKED ISSUE #797: Element not recognized for atom name 0C21 1C21
Hi,
Found this atom names inside a pdb of a lipid bilayer. NGL leaves empty the element property of the atom and breaks bonds.
Maybe it is due to the first character being a number?
Attached is the pdb file. 

[POPC_303K.pdb.txt](https://github.com/nglviewer/ngl/files/5288552/POPC_303K.pdb.txt)

fredludlow:
As per the title - fix the check for first char of atomname is integer in the `guessElement` function.


garyo (approved):
LGTM!

PR #836, opened 2021-01-23
Resolve #829

### LINKED ISSUE #829: "o.addRepresentation('surface' " has a bug with cube format files
I am developing a custom JupyterLab extension to visualizer cube format file. There is a problem when using
the "o.addRepresentation('surface' " with cube format files. It gives error as:

**Uncaught SyntaxError: Function statements require a function name**

This problem is only for cube format file. If I use other format file to create 'surface', it will not cause
any error. Could you give any suggest which function could lead to this error ? Thanks.

dou-du:
The function without a return type will cause a problem (function without a function name). I simply add
a return type as any, which fixes my issue.

### DIFF
--- a/src/surface/marching-cubes.ts
+++ b/src/surface/marching-cubes.ts
@@ -322,7 +322,7 @@ function getTriTable (): Int32Array {
 // are drawn as they're redundant
 // The line between edge 1 and 5 is always drawn as it's on the leading edge
 
-function getAllowedContours () {
+function getAllowedContours (): any {
   return [
 
     [ 0, 4, 4, 4, 2, 0, 0, 0, 2, 2, 0, 0 ], // 1 2 3 4 8 9


fredludlow:
Thanks for the PR @dou-du - I wonder if putting the correct return type would be better? (I think it's `Array<number[]>` but not tested)

dou-du:
> Thanks for the PR @dou-du - I wonder if putting the correct return type would be better? (I think it's `Array<number[]>` but not tested)

The Array<number[]> will lead to error:

```bash
The left-hand side of an arithmetic operation must be of type 'any', 'number', 'bigint' or an enum type.

744         if (allowedContours[ e2 ][ e3 ] & edgeFilter) {

```

fredludlow:
I'm a bit confused as to why that would be - even without any return type specified for that function typescript infers the correct type (`number[][]`, which is neater than `Array<number[]> anyway`) and compiles without any complaints.

You mentioned webpack, is webpack reading the typescript files directly rather than the built module? (NGL uses rollup to build a module)
And if so, are you doing it that way because we don't currently export type information?
And if so... would exporting types properly be a better way to fix this? (I know @ppillot and @garyo did a lot of work towards this and the type declarations didn't make it into the NPM package yet, but it would be great if we could do that).

To come back to the original PR, I'm sort of okay with adding this as a quick fix, but I'd like to understand how you're using webpack etc and whether there's a more fundamental fix (exporting type declarations or other) that would make this and related problems go away.

garyo:
I agree with Fred here; Typescript is correctly determining the return type of `getAllowedContours` already (it's `number[][]`). I think your actual problem is likely elsewhere.

ppillot:
> And if so... would exporting types properly be a better way to fix this? (I know @ppillot and @garyo did a lot of work towards this and the type declarations didn't make it into the NPM package yet, but it would be great if we could do that).

Yes, that would be the best way to do this. I havent' had time to contribute recently, but I use a completely typed version of NGL which plays nicely with other TS code. Shipping declarations with the npm bundle would be great.

Issue #838, opened 2021-01-26
Download Selection of Structure


waltonseymour:
It would be useful to download/write just a selection of a structure for the `Writer` classes

PR #840, opened 2021-02-01
Enables writing partial structures (for PDB format) as requested in #838

fredludlow:
Allows partial structures to be written (by passing the appropriate `StructureView` object to PdbWriter as opposed to a `Structure` object).

e.g, given a structure `struc` you can write the data for atoms '1-10':

```javascript
var sv = struc.getView(new NGL.Selection('1-10'))
var writer = new NGL.PdbWriter(sv)
writer.getData() // Text for atoms in residues 1-10
```

### DIFF
--- a/src/writer/pdb-writer.ts
+++ b/src/writer/pdb-writer.ts
@@ -118,7 +118,7 @@ export default class PdbWriter extends Writer {
           defaults(a.element, '')
         ))
         ia += 1
-      })
+      }, this.structure.getSelection() || undefined)
 
       this._records.push(sprintf('%-80s', 'ENDMDL'))
       im += 1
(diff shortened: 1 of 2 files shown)


src/writer/pdb-writer.ts
```diff
@@ -118,7 +118,7 @@ export default class PdbWriter extends Writer {
           defaults(a.element, '')
         ))
         ia += 1
-      })
+      }, this.structure.getSelection() || undefined)
```
waltonseymour, line 121: Would it make sense to have `getSelection()` default to `undefined` rather than `false` to simplify this?
This looks great btw!

garyo:
Looks good to me; nice & simple.

garyo (approved):
LGTM

src/writer/pdb-writer.ts
fredludlow: Yes :) Fixed in another commit

Issue #800, opened 2020-10-01
SAS surface and SES surface look the same


DAlvGar:
Hi all,

I detected a potential bug. SES surface (Solvent excluded surface) looks exactly the same as the SAS (solvent accessible surface). Reproducible in http://nglviewer.org/ngl/ molecular surface example.
They should be different: https://www.cgl.ucsf.edu/chimerax/docs/user/commands/surface.html

How could I fix this? 

Thank you all!

Dani

fredludlow:
So I think the bug is in SES. It should look approximately the same as tehe av method (but calculated differently)...

DAlvGar:
Hi Fred, thanks for checking this. Is anything I could help with? I took a look to the calcualtion but, being honest, I am a bit lost in there to know what to modify, any tips would be helpful.

fredludlow:
I just had another quick look and I'm not really any the wiser. 

I checked out version 0.9.0-dev.0 (the oldest tag I could easily build) just in case it was a regression during some of the typescript conversion or in case I'd broken marching cubes (I made a few changes when adding the 'av' surface type) - anyway, the behaviour is the same there too, so whatever it is it's been like that a while. 

https://fredludlow.com/ngl/v0.9.0-dev.0/examples/webapp.html

How badly do you want SAS? (I think SES and AV are equivalent (and AV looks nicer)). From that diagram I guess it's "just" a case of following the VWS code and adding a constant to all the atom radii...

Sorry, that's probably not super helpful...

PR #843, opened 2021-02-16
Update defaults: set surface type to vws, increase surface resolution

garyo:
This just updates a couple of defaults to be slightly nicer -- the default surface resolution is very low, which is good for performance, but I think modern systems can easily handle more triangles so why not make it look better by default? 
And it seems to me like a zero-offset van der Waals surface is a better surface default than the current 1.4Å-offset surface-accessible surface, but I'm happy to remove that if people prefer the current one.

### DIFF
--- a/src/stage/stage.ts
+++ b/src/stage/stage.ts
@@ -351,13 +351,13 @@ class Stage {
             )
           )
         )
-        if (backboneOnly) scaleFactor = Math.min(scaleFactor, 0.15)
+        if (backboneOnly) scaleFactor = Math.min(scaleFactor, 0.5)
 
         component.addRepresentation('surface', {
           colorScheme, colorScale, colorReverse,
           sele: 'polymer',
-          surfaceType: 'vws',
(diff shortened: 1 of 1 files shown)


fredludlow:
> This just updates a couple of defaults to be slightly nicer -- the default surface resolution is very low, which is good for performance, but I think modern systems can easily handle more triangles so why not make it look better by default?

Agreed

> And it seems to me like a zero-offset van der Waals surface is a better surface default than the current 1.4Å-offset surface-accessible surface, but I'm happy to remove that if people prefer the current one.

I'd go for 'av' I think it looks nicer (I'm biased here, so feel free to overrule this). The 'av' surface is also a Connolly surface (i.e. solvent-excluded surface, though the builtin 'ses' surface isn't..  - see #800 ) which has the nice property that protein and ligand surfaces generally 'mesh' quite nicely at a probe-size of about 1.4-1.5, so for visual inspection of how well a ligand fits a binding site it's a good choice.

fredludlow (approved):

PR #844, opened 2021-02-16
Update build system to latest rollup for better sourcemaps, plus build typescript decls

garyo:
This updates to the latest rollup, with typescript plugin so rollup does the typescript transpilation. This makes for much better sourcemaps, as well as a cleaner build. I tried to clean up the rollup config while doing this -- I believe I got all three builds correct, but I don't have a way to test them all. I use the ESM one so I can confirm that's working properly.

Also builds the typescript declarations into the distribution (finally).


garyo:
Note that the typescript declarations are created in `dist/declarations`; that seems like a good place for them but I'm open to any other location.

fredludlow (approved):
LGTM, much cleaner.

fredludlow:
Thanks for this @garyo - you've marked it as draft so I've not merged it but go ahead if you're happy with it.

garyo:
Thanks -- with f07545f the bundled build works as well as ESM for me. We don't have any tests for the UMD build though, right? Not clear if anyone cares about that anymore anyway...

PR #845, opened 2021-02-17
Add a few more typescript decls

garyo:
There are certainly more that could be added but these are a start.


fredludlow (approved):

fredludlow:
Thanks @garyo!

PR #847, opened 2021-02-22
Include Typescript declarations in published npm package

sto3psl:
I noticed while using `ngl` in a Typescript project, that the Typescript compiler couldn't find `ngl`s types. After some digging I noticed that the `ngl.d.ts` file was correctly referenced in `package.json` but `tsc` complained. 

Then I checked inside the `declarations/` folder in my `node_modules/ngl` only to notice, that the folder didn't exist. This meant types can get generated but they are not published to npm. 

To fix this I added `declarations` to the `files` field in `package.json` since only the files defined there will end up published on npm. -> https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files

I'm not familiar with how `ngl` is published so maybe we need to update some pre publish scripts (?).

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,8 @@
     "url": "https://github.com/arose/ngl/issues"
   },
   "files": [
-    "dist"
+    "dist",
+    "declarations"
   ],
   "author": "Alexander Rose <alexander.rose@weirdbyte.de>",
   "contributors": [],


garyo:
Yes, you're right -- I have a pending PR to fix this at least partially. I'm not sure it uses prepublish; I'll have to check.

sto3psl:
Is it this one #844? Moving declarations to the `dist` folder works too. Since that fixes the issue too, feel free to close this PR.

sto3psl:
@garyo I saw that you merged #844 with the declarations fix. It would be great, if you could publish a new release to npm for people to use. Thanks!

PR #849, opened 2021-03-03
Catch errors in sprintf

fredludlow:
Logs a useful error message to the console when your `labelFormat` string is broken, rather than just failing


garyo (approved):
LGTM!

PR #854, opened 2021-03-20
Add CallbackTrajectory

yitzchak:
This PR adds a new trajectory type that gets the frame count and individual frames from a callback function. This functions like `RemoteTrajectory` but leaves details of the request mechanism up to the callback function. 

In order to request the frame count the function is called with a single argument which is a callback that accepts the frame count as its first argument.

To request an individual frame the function is called  with a callback, the frame number and the atom indices. The callback takes 4 arguments which are the frame number, the box, the frame coordinates, and an updated frame count. 

Both of these requests mirror the behavior of the http requests of `RemoteTrajectory`.

We use ngl extensively in our cando-jupyter kernel. In the past we have reused the front end code of nglview and implemented a Common Lisp backend that runs in Cando. We are currently making a transition to our own Jupyter frontend code in [ngl-clj](https://github.com/yitzchak/ngl-clj/tree/shasht) which has a more lispy object oriented interface among other things.

In order to reuse the trajectory logic in ngl (frame caching, interpolation, etc.) we are implementing a trajectory server using COMM remote calls in the Jupyter protocol. One obstacle to this is that none of the trajectory classes are exported so we cannot create a new subclass easily. 

Adding this new trajectory type enables custom remote style trajectories (like our Jupyter protocol) to be implemented while still keeping the implementation details of ngl private.

Also, since the name of this trajectory type  probably isn't very obvious a better one could certainly be chosen.

FYI @drmeister @myonkunas

### DIFF
--- a/examples/scripts/interactive/callback-trajectory.js
+++ b/examples/scripts/interactive/callback-trajectory.js
@@ -1,0 +1,20 @@
+
+stage.loadFile('data://ala3.pdb').then(function (o) {
+  let position = o.structure.getAtomData({ what: { position: true } }).position
+  let count = 100
+  let radius = 0.1
+
+  o.addRepresentation('licorice')
+  o.addTrajectory(function (cb, i, atomIndices) {
+    if (typeof i === 'number') { // This is a frame request
+      let frame = Float32Array.from(position)
+      for (let j = 0; j < frame.length; j++) { // Add random displacement to positions
+        frame[j] += (2 * Math.random() - 1) * radius
+      }
+      cb(i, null, frame, count) // return the new frame
+    } else { // This is a frame count request
+      cb(count)
+    }
+  })
+  o.autoView()
+})
--- a/src/trajectory/callback-trajectory.ts
+++ b/src/trajectory/callback-trajectory.ts
@@ -1,0 +1,72 @@
+/**
+ * @file Callback Trajectory
+ * @author Tarn W. Burton <twburton@gmail.com>
+ * @private
+ */
+
+import Structure from '../structure/structure'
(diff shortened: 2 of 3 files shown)


fredludlow:
Hi @yitzchak, thanks for this, the ngl-clj project looks interesting! It looks fine to me and I'm happy to merge this more or less as-is (very minor comment is maybe put your own details in the @author line? (or your github @yitzchak handle if you'd rather keep email address private) as the author)?

Also, this isn't critical to merging but if you could add an example script that would be really handy - I *think* I can imagine what's going on but I don't have any way to test it myself :)

yitzchak:
Thanks for the feedback, I'll make the changes and update. 

One other possible is to call it a `CallbackTrajectory` instead of `FunctionTrajectory`. Thoughts?

fredludlow:
+1 for `CallbackTrajectory`

yitzchak:
Renamed to `CallbackTrajectory` and added a simple example to the webapp. Should be good to go.

fredludlow (approved):
Looks good, and thanks for the example!

fredludlow:
Thanks @yitzchak, looks great!

yitzchak:
Thanks for the quick turnaround. Do you have any idea on a timeline for a npm publish? We can link to git revision until then, but just curious.

fredludlow:
I have just published 2.0.0-dev.39, which you should be able to get by `npm i ngl@next`

(I'm still trying to iron out some of the wrinkles in the scripts for doing npm releases so let me know if it works or if I broke anything)

yitzchak:
Awesome. Thank! I'll let you if there are issues.

PR #866, opened 2021-05-07
Fix volume surface gen when changing isolevel so geometry gets smaller

garyo:
In this case, instead of creating a new `index` attribute array, it reuses the old one, which is fine, but it didn't reset the `count`, which could cause the old triangles to be reused, causing graphics glitches. Simple fix (but it took me a long time to find it!)


fredludlow (approved):

fredludlow:
Thanks @garyo, that does look well hidden :)

PR #870, opened 2021-06-08
Possible fix for interactive/crosslinking example on sandbox

cmdcolin:
Currently the interactive/crosslinking example codesandbox is a blank screen

It has nested double quotes in a string, which is a syntax error. This PR proposes changing it to use a template string

Not clear why it was changed to double strings in the code sandbox or why it didn't escape the double quotes, but I thought using template strings may be a workaround


cmdcolin:
Also thanks for this library...currently investigating it's usage and really like all the example codepens!

garyo (approved):
Looks good to me!

PR #882, opened 2021-09-20
Housekeeping

fredludlow:
I've tried to do a bit of cleanup of the dependencies and build process, some of this is uncontroversial bumping versions / removing deprecated plugins, and some of it changes build behaviour a bit, so please comment if you have a view!

I've updated the deprecated rollup plugins (typescript, buble etc).

The build-min job has changed quite a lot, it used to take the uncompressed dev build form the build/ folder and run it through uglify-js to create dist/ngl.js. It would also copy the esm and umd modules to dist too. 

I now use rollup with rollup-plugin-tersify (removing uglify-js dependency) so the minified versions are built in a rollup job from the source .ts files. The main advantage of this is that correct sourceMaps can be generated for the minified files. **I have also changed the sourceMaps so that they include the (typescript) sources (in the .map.js file)**. This makes the map files bigger but it means you can easily debug NGL issues from any deployment. I can make this a separate option if anyone dislikes the larger map files. 

Finally, I've removed eslint as an explicit dependency (we're not currently using it) - though I think in practice it gets pulled in by something else anyway.

In a future PR, it would be good to sort out standard (prettier?) and eslint for autoformatting and linting. I played with eslint but we have a huge number of warnings and errors at the moment and I didn't really want to turn on eslint only to disable most of the rules... (My progress on this is in the fredludlow/ngl/eslint branch for reference, but I'm not working on it right now)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,511 +5,490 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.10.4",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.10.4.tgz",
-      "integrity": "sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==",
+      "version": "7.14.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.14.5.tgz",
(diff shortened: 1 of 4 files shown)


fredludlow:
@garyo, @ppillot  - any thoughts? If not I'll merge this

ppillot:
Thanks @fredludlow I am looking at it and will keep you posted quickly.

ppillot (approved):
Works as advertised. Helpful simplification. Thanks!

ppillot:
Tests are successful. Thanks!

garyo:
This looks great; thanks @fredludlow !

fredludlow:
Thank you both!

PR #887, opened 2021-09-30
Fix for cylinder impostors in orthographic code

### LINKED ISSUE #883: Broken capping for impostor cylinders in orthographic camera mode
Hi,
I noticed there is a broken capping for impostor cylinders in orthographic camera mode 
Orthographic mode (Broken)
image

Perspective mode (OK)
image

fredludlow:
This should fix the bug @giagitom reported (closes #883) with missing cylinder caps (and interiors) in orthographic mode.

The diff is bigger than it needs to be as it took me a while and a few comments to grok what was going on, and I thought I might as well leave the comments in (and simplify the code in a few places) for when I inevitably have to look at it again :)

The critical change is removal of line: https://github.com/nglviewer/ngl/blob/3bc074b027a07b065b2dcb8da1083fff14004b3c/src/shader/CylinderImpostor.frag#L131

### DIFF
--- a/src/buffer/mappedalignedbox-buffer.ts
+++ b/src/buffer/mappedalignedbox-buffer.ts
@@ -7,6 +7,17 @@
 import { BufferParameters, BufferData } from './buffer'
 import MappedBuffer from './mapped-buffer'
 
+//       +Y   /
+//    0**********2
+//    *   | /   **
+//    *   |/   * *
+// -----------3---- +X
+//    *  /|   *  *
+//    * / |   *  *
+//    1/**|******4
+//    /   |   * *
+//   /    |   **  
+//  +Z    |   5 
 const mapping = new Float32Array([
   -1.0, 1.0, -1.0,
   -1.0, -1.0, -1.0,
--- a/src/shader/CylinderImpostor.frag
+++ b/src/shader/CylinderImpostor.frag
@@ -82,26 +82,33 @@ float calcClip( vec3 cameraPos ){
 
 void main(){
 
-    vec3 point = w.xyz / w.w;
+    // The coordinates of the fragment, somewhere on the aligned mapped box
+    vec3 ray_target = w.xyz / w.w;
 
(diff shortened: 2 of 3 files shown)


giagitom (commented):
Hi @fredludlow ,
Thank you very much for this fix! Orthographic mode works nicely now. 
However I notice a strange artifact in perspective camera mode having cylinder perpendicular to camera. Something that should be discard i think
image
with a different angle this can't be seen
image

fredludlow:
Good spot! I *think* I know why that is, will try and fix it.

fredludlow:
I think the commit I just pushed should fix this (and tidies up a few other bits) - could you take a look please @giagitom?

I wonder if it's also sensible to make impostors front-side only by default? The only difference I can see is a slight artefact in double-sided mode + perspective rendering + opacity != 1.0:

image

(The darker region is because you're seeing both the side of the aligned box and the inside/back-face of the near end (due to the camera being really close)

giagitom:
Thank you @fredludlow , the artifact is gone now! This is more than enough for me!

Probably are still present some little issues when camera is close enough. Sometimes cylinders disappear in strange ways.

for example this
image


or this 
image
when zooming in a little bit it suddenly became this
image

fredludlow:
Yeah, there's definitely some oddities at very close zoom - I don't think I've introduced any of these with the changes in this PR (looking at http://nglviewer.org/ngl/?script=showcase/ferredoxin I see similar things), but if there are some that are a result of these changes please let me know!

I'm less sure about how to fix the ones above, I guess they're when the camera is inside the aligned box?

giagitom:
Ohh.. yes sorry. This wasn't an issue with the current PR. 

 > I'm less sure about how to fix the ones above, I guess they're when the camera is inside the aligned box?
Yes i think, maybe it is something that can't be fixed?

giagitom (approved):

Issue #892, opened 2021-11-24
Is there a way to add new attributes to atom? 


appassionate:
Hi, I want to present my data in every atom
I imaged a customized colorscheme will read the value of  atoms if i can do this successfully.

It might exist in NGL:   ```atom.resno```
can I get ```atoms._input_value``` in NGL?
Thanks!

ppillot:
The list of fields in the atomStore is preconfigured.
To achieve a coloring per external property as you are mentioning, I've created a colorMaker that polls values from a separate data structure.

```typescript
    class MyColorMaker extends NGL.Colormaker {
            myScale: ColormakerScale;

            constructor (params: any) {
                super(params);
                this.parameters.scale = 'OrRd';
                this.parameters.domain = [lMin, lMax];
                this.myScale = this.getScale();
            }
            atomColor (a: AtomProxy) {
                const val =myDataStore[a.serial];
                return (val !== -1) ? this.myScale(val) : 0xcccccc;
            }
    }

        NGLColormakerRegistry.add('mycolor', myColorMaker);

```

ppillot:
The list of fields is preconfigured.
What you can do is create a ColorMaker that will poll values from your own data structure based on the atom id.

appassionate:
> The list of fields is preconfigured. What you can do is create a ColorMaker that will poll values from your own data structure based on the atom id. https://nglviewer.org/ngl/api/manual/coloring.html#custom-coloring

thanks! actually, i'm nglview user. Maybe it will be easier to add my data using this function, but in python console i have no idea about to trans by numpy numerical array to a data in a TS or JS form...
A trick thinking is that if i can update specific color in an exist representation just as adding a reprensentation by ```sele= "@1","@2","@3"```. Or a presentation just owns a colorscheme?  thank you : )

hainm:
@appassionate maybe try to follow those threads in `nglview`:
- https://github.com/nglviewer/nglview/issues/913
- https://github.com/nglviewer/nglview/pull/839

fredludlow:
I do a similar thing (more or less per @ppillot's link, custom class polls data from another source). This gets messier when:

1. The source is async
2. Serializing state (persisting parameters etc) - requires also serializing/reproducing state of the data source

I've been toying with a few ways around this and wondered whether having a general method for passing data through to custom colormakers would be useful? I think it would make a few things I do less verbose as no custom colormaker would be required, just some paramters.

```typescript
[...]

PR #897, opened 2021-12-15
Colordata

fredludlow:
This is partly in response to #892,  Adds a `colorData` parameter to representations, that colormakers may make use of, and a new builtin `structuredata` colormaker will make use of this where colorData is of shape 

```
{ atomData?: { [atomIndex: number]: number}, 
  bondData?: { [bondIndex: number]: number}
}
```

See the examples color/atom-data and color/bond-data for usage.

I'm open to alternative ways to achieve this. One obvious alternative would be to allow arbitrary (atom-based?) data to be added to `Structure` objects. The main reason I didn't do this is that I couldn't think of a clean way to connect updates in the extra data to updating coloring of representations without some objects knowing more about others than they ought to. (If `Structure` sends a signal saying that indicates the data with attribute name "_some_fancy_model_output" has been updated, should a representation trigger a re-color?)

Note that at runtime there's no check that you're passing in colorData of the shape described above, so it could be used to pass any data through to your custom colormaker - if that's preferable as a general method we could update the typing info to support it.

Comments (major changes as well as naming/nitpick etc) very welcome!


src/color/colormaker.ts
```diff
@@ -46,10 +46,16 @@ export const ScaleDefaultParameters = {
 }
 export type ScaleParameters = typeof ScaleDefaultParameters
 
+export interface ColorData {
+  atomData?: { [key: number]: number }
```
ppillot, line 50: This type definition is an inlined keyed object, when we are using number arrays.
I think it would be better typed by the following
```typescript
export interface ColorData {
    atomData?: number[],
    bondData?: number[]
}
```

examples/scripts/color/atom-data.js
```diff
@@ -0,0 +1,29 @@
...
+  var repr = c.addRepresentation('ball+stick', {
+    color: 'structuredata',
+    colorData: {atomData: distance},
+    colorDomain: [30.0, 60.0]
+  })
+  c.autoView()
+
+  // If colorData isn't defined, empty etc, falls back to value, here we wait 2 seconds
+  // then update the paramter to an empty array to demonstrate
+  window.setTimeout(() => {
```
ppillot, line 26: Not sure if it's a good idea to remove a coloring when demonstrating the feature. It might give a wrong impression.

src/color/structuredata-colormaker.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+import { ColormakerRegistry } from '../globals'
+import Colormaker, { ColormakerScale, manageColor, StuctureColormakerParams } from './colormaker'
+import AtomProxy from '../proxy/atom-proxy'
+import BondProxy from '../proxy/bond-proxy'
+
+
+class StructuredataColormaker extends Colormaker {
+  atomData?: { [key: number]: number }
```
ppillot, line 14: same as above, expected format is `number[]`. For consistency we could also import the `ColorData` interface here and define `atomData` type as `ColorData['atomData']`

ppillot (commented):
These are minor suggestions.

Maybe a major concern would be about the expected format as using an array should be favored to an inlined object.

examples/scripts/color/atom-data.js
fredludlow: Yeah, I agree, I was being lazy - the idea was to show that the color falls back to `colorValue` if data is missing from the input, but a better way is probably to just provide some partial data in the first instance - will change this.

src/color/structuredata-colormaker.ts
```diff
@@ -34,7 +34,18 @@ class StructuredataColormaker extends Colormaker {
   @manageColor
   bondColor(bond: BondProxy, fromTo: boolean) {
       const val = this.bondData?.[bond.index]
-      return (val !== undefined) ?  this.scale(val) : this.parameters.value
+      
+      // Explicit bond data?
+      if (val !== undefined) return  this.scale(val)
+      
+      // No, fall back to atomData
```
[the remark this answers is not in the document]
fredludlow: Will update example as per the other comment..

src/color/colormaker.ts
fredludlow: Also agreed (I was trying to keep it general for cases where you want to only supply data for a subset of atoms - i.e. something like: `atomData:  {0: 0.5, 55: 2.3, 148: 1.1}` but this is probably just premature optimization - I think under the hood sparse arrays are stored fairly efficiently in JS).

fredludlow:
Thank you very much @ppillot for the review! I think all the issues you raised are fixed in 7505d9e, plus a bug I'd overlooked with how bondColor interacts with the manageColor decorator.

ppillot:
I'll have a look when I can. Thanks for the hard work @fredludlow!

ppillot (approved):
Thanks! Great additions, and  clear examples.
Tested and approved

Issue #696, opened 2019-09-13
load mmCIF file error


lidaof:
Hi there,

I am trying to load a mmCIF format file using ngl from http://nglviewer.org/ngl/?script=showcase/ferredoxin, got error:
```
ngl.js:1 STAGE LOG error loading file: 'TypeError: Cannot read property 'charCodeAt' of undefined'
?script=showcase/ferredoxin:1 Uncaught (in promise) error loading file: 'TypeError: Cannot read property 'charCodeAt' of undefined'
```
My file is [here](https://wangftp.wustl.edu/~dli/tmp/GSM3271347_gm12878_01.impute3.round4.clean.3dg.cif). Same file can be load by [3dmol](http://3dmol.csb.pitt.edu/doc/index.html) or pymol2, not sure what  is the format error? Thanks!

fredludlow:
First off - I'm really not familiar with the cif format (outside of the fairly standardised PDB version).

The cif-parser expects altloc to be one of your columns in the cif file.

`atomStore.altloc[idx] = altloc.charCodeAt(0);`

This is easy to fix (just check for undefined)

I started down this route here:
https://github.com/fredludlow/ngl/tree/696-cif

but there are other issues - this cif-parser expects certain names (eg `ptnr1_auth_seq_id`) but the example you've got calls it `ptnr1_label_seq_id`. Is there a definitive set of acceptable names somewhere? Do pymol and 3dmol parse the bonding section?

lidaof:
Hi @fredludlow thank you for looking into this.
I am actually pretty new to this field, was trying to visualize this cif format using ngl. I used [dip-c](https://github.com/tanlongzhi/dip-c#interactive-visualization-of-3d-genomes) to generate the file. I will also look at the cif format specification. Thanks again.

lidaof:
@fredludlow also  `ptnr1_label_seq_id` seems like a valid name as : http://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_struct_conn.ptnr1_label_seq_id.html

fredludlow:
I guess we need to figure out why NGL looks for `ptnr1_auth_seq_id` rather than `ptnr1_label_seq_id` and whether that's: just a bug; a better thing to do when it's available or the required behaviour for some other non-obvious reason :)

Is anyone with a good knowledge of mmcif format reading? :)

arose:
[...]

PR #903, opened 2022-01-20
Use label_seq_id if auth_seq_id is not present in CIF

luxaritas:
Per https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.auth_seq_id.html, auth_seq_id is an alternative numbering to match a publication, whereas label_seq_id is a sequential identifier. It is possible to have CIF files with the latter, but not the former (in my case, the file I was using was created by extracting the chain chain LB from PDB 3J9Z in PyMOL, then exported to CIF).

Related: #696

### DIFF
--- a/src/parser/cif-parser.ts
+++ b/src/parser/cif-parser.ts
@@ -760,7 +760,7 @@ class CifParser extends StructureParser {
     var first: boolean|null = null
     var pointerNames: string[] = []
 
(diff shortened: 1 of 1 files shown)


fredludlow:
Thank you @luxaritas - looks good to me!

fredludlow (approved):

PR #905, opened 2022-01-21
Export representation classes to allow for extension

luxaritas:
I'm creating a custom StructureRepresentation for my embedded NGL viewer, where I want to extend the StructureRepresentation class. I've gone ahead and exported all representations and their parameters as well, as if someone wants to make a representation based on an existing representation, those are also exposed.

### DIFF
--- a/src/ngl.ts
+++ b/src/ngl.ts
@@ -87,31 +87,32 @@ import './component/volume-component'
 
 //
 
-import './representation/angle-representation'
-import './representation/axes-representation'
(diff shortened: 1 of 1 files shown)


fredludlow:
Thanks @luxaritas. This looks fine, I guess the only slight concern is that the public NGL namespace is getting big, but I'm not sure a big reorganization of that would help anyone right now. @ppillot and @garyo any comments?

PR #906, opened 2022-01-21
Improve typing of "impostorable" buffers

luxaritas:
This PR allows instances of CylinderBuffer to be properly be typed as a `CylinderGeometryBuffer | CylinderImpostorBuffer`, as that's what the constructor returns.

FWIW: I'm not a fan of the `as any` casting as it avoids proper type checking. It might be preferable to change the API here to be a static method instead.

### DIFF
--- a/src/buffer/cylinder-buffer.ts
+++ b/src/buffer/cylinder-buffer.ts
@@ -38,7 +38,7 @@ export type CylinderBufferParameters = typeof CylinderBufferDefaultParameters
  *   radius: new Float32Array([ 1 ])
  * });
  */
-class CylinderBuffer {
+class CylinderBufferImpl {
   constructor (data: CylinderBufferData, params: Partial<CylinderBufferParameters> = {}) {
     if (!data.color2 && data.color) data.color2 = data.color
     if (!ExtensionFragDepth || (params && params.disableImpostor)) {
@@ -48,6 +48,12 @@ class CylinderBuffer {
     }
   }
 }
+
+const CylinderBuffer: {
+  new(data: CylinderBufferData, params: Partial<CylinderBufferParameters>): CylinderGeometryBuffer | CylinderImpostorBuffer;
+} = CylinderBufferImpl as any;
+
+type CylinderBuffer = CylinderGeometryBuffer | CylinderImpostorBuffer;
 
 BufferRegistry.add('cylinder', CylinderBuffer)
 
--- a/src/buffer/cylinder-buffer.ts
+++ b/src/buffer/cylinder-buffer.ts
@@ -23,6 +23,17 @@ export const CylinderBufferDefaultParameters = Object.assign({
 }, CylinderGeometryBufferDefaultParameters, CylinderImpostorBufferDefaultParameters)
 export type CylinderBufferParameters = typeof CylinderBufferDefaultParameters
 
+class CylinderBufferImpl {
+  constructor (data: CylinderBufferData, params: Partial<CylinderBufferParameters> = {}) {
+    if (!data.color2 && data.color) data.color2 = data.color
+    if (!ExtensionFragDepth || (params && params.disableImpostor)) {
+      return new CylinderGeometryBuffer(data, params)
+    } else {
+      return new CylinderImpostorBuffer(data, params)
+    }
+  }
+}
+
 /**
  * Cylinder buffer. Depending on the value {@link ExtensionFragDepth} and
  * `params.disableImpostor` the constructor returns either a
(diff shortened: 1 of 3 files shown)


luxaritas:
Realized the same thing applies to hyperballstick and sphere, so adjusted those as well

ppillot:
Thanks for noticing this.
Could you describe, (the best would be a code example), the issue you were encountering?
It looks to me that adding a supplementary dynamic constructor step in the code to solve an issue with static typings is adding a level of complexity that we may be able to avoid by other means (improving the static typings).

luxaritas:
The example would be if you have something like:

`const buff = new CylinderBuffer(..., ...)`

`buff` in this case would have no methods or attributes, as all the `CylinderBuffer` class has is a constructor - even though the constructor actually returns a `CylinderGeometryBuffer | CylinderImpostorBuffer`.

TypeScript doesn't allow for specifying a return type on a constructor, so the only alternative to this is creating a static method like `CylinderBuffer.create()` which _could_ be properly typed. The reason I didn't do that is because that's a breaking API change - however I'd argue that's probably a more idiomatic way of doing it.

ppillot:
Thanks for the clarification. I've played a bit with this and it looks like there are several problems here with regard to modelling this constructor in TS: parameters that are marginally incompatible, a conditional in the constructor that TS can't model.... I have tried to use the "old" JS objects constructor syntax (using `function` instead of `class`) but TS still chokes on types compatibility and it does not really improve the code.

src/buffer/cylinder-buffer.ts
```diff
@@ -49,6 +49,12 @@ class CylinderBuffer {
   }
 }
 
+const CylinderBuffer: {
+  new(data: CylinderBufferData, params: Partial<CylinderBufferParameters>): CylinderGeometryBuffer | CylinderImpostorBuffer;
+} = CylinderBufferImpl as any;
+
+type CylinderBuffer = CylinderGeometryBuffer | CylinderImpostorBuffer;
```
ppillot, line 56: In my tests, this type declaration did not seem necessary. Also, it might cause some confusion with the class wrapper name.

src/buffer/cylinder-buffer.ts
```diff
@@ -49,6 +49,12 @@ class CylinderBuffer {
   }
 }
 
+const CylinderBuffer: {
+  new(data: CylinderBufferData, params: Partial<CylinderBufferParameters>): CylinderGeometryBuffer | CylinderImpostorBuffer;
+} = CylinderBufferImpl as any;
```
ppillot, line 54: Instead of using the `any` type assertion, I would suggest to use a `//@ts-expect-error` comment. For example `//@ts-expect-error Incompatible parameters types between constructors` before the declaration to document the issue and maybe have it resolved through future evolutions of TS.

ppillot: Also, the informations provided by the JS-doc comment are now disconnected from CylinderBuffer. Please, move them before the `const` declaration.

ppillot (changes requested):

src/buffer/cylinder-buffer.ts
luxaritas: `CylinderBuffer` is used as a type in a number of places - for example, `UnitcellRepresentation#cylinderBuffer` is typed as `CylinderBuffer`. If you prefer, I can change all instances of the CylinderBuffer type to be `CylinderGeometryBuffer | CylinderImpostorBuffer`, but this just seemed to me to be a handy shortcut (especially as the idea behind this is that you think of it as a CylinderBuffer in general, and don't care whether it's a GeometryBuffer or an ImpostorBuffer)

src/buffer/cylinder-buffer.ts
luxaritas: Sure, makes sense

src/buffer/cylinder-buffer.ts
luxaritas: @ppillot Any update on this?

ppillot: Sorry!
I am still feeling queasy about this. I can see that the PR works. It is unclear to me though how the 2 symbols, one for typings, the other for a constructor, are coexisting. There is some documentation in typescript reference [about declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) but the specific case presented here is not specified (merging const name and type name). I can't tell if this is guaranteed to work in the future.
Overall I am indecisive regarding this PR. I won't object for it to be merged. @garyo @fredludlow any opinions?

luxaritas: `const CylinderBuffer` is not a type, it is a runtime value. Similarly `type CylinderBuffer` is a type, not a runtime value. AFAIK, you are allowed to define type and runtime symbols with different names, and a type is only implicitly created for a runtime value in the case of a class, enum, etc.

I get your concern though - as I said, if the preference is to change all `CylinderBuffer`s to `CylinderGeometryBuffer | CylinderImpostorBuffer`, we can do that.

luxaritas: Oh I see what you're saying wrt those docs, I misread your comment. Yeah my understanding is that TS treats type names and value names as separate sets of symbols, and creates them according to the rules given in the table.

luxaritas: Let me know whether the desired outcome is preserving `CylinderBuffer` or changing usages of that `CylinderGeometryBuffer | CylinderImpostorBuffer`, and I can make any changes as needed

papillot (approved):

ppillot (approved):

PR #907, opened 2022-01-25
Update type of BondData#picking to the more correct BondPicker

luxaritas:
This parameter is always set as a BondPicker, so I've typed it accordingly in order to make the relevant methods available

### DIFF
--- a/src/structure/structure-data.ts
+++ b/src/structure/structure-data.ts
@@ -1,6 +1,6 @@
 
 import BitArray from '../utils/bitarray'
-import { AtomPicker } from '../utils/picker'
+import { AtomPicker, BondPicker } from '../utils/picker'
 import BondStore from '../store/bond-store'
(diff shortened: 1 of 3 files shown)


fredludlow:
Thanks @luxaritas for this - this allows a few other things to be simplified (I pushed a new commit to your branch) - could you sanity check those for me and I'll merge it. I've tested it merged into master locally as well and it's fine with new TS

fredludlow:
Thanks @luxaritas for this - this allows a few other things to be simplified (I pushed a new commit to your branch) - could you sanity check those for me and I'll merge it. I've tested it merged into master locally as well and it's fine with new TS

fredludlow (approved):

luxaritas:
The `as any` in `bondData.picking = new BondPicker(...) as any` is no longer needed now, right? Otherwise, looks reasonable

PR #908, opened 2022-01-26
Fix crash when using alternate materials (eg when using outline effect)

luxaritas:
In my app, I've integrated rendering using EffectComposer, including an OutlinePass. The OutlinePass relies on rendering with special materials, which are not `ShaderMaterial`s and as such do not have the uniforms property.

Let me know if there's a better way to handle this (and if I should change the type of `material` to `Material`)

### DIFF
--- a/src/viewer/viewer.ts
+++ b/src/viewer/viewer.ts
@@ -58,6 +58,8 @@ function onBeforeRender (this: Object3D, renderer: WebGLRenderer, scene: Scene, 
   const u = material.uniforms
   const updateList = []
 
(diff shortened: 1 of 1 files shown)


fredludlow (approved):

luxaritas:
I don't have a minimal reproduction - let me know if you'd like one. I use it in context [here](https://github.com/eternagame/EternaJS/blob/bc4d2a6f1d859c7184eb368eba7e08ccaa3c33c1/src/eterna/pose3D/Pose3D.ts#L189)

PR #909, opened 2022-01-26
Feature/rotate selection and pivot

garyo:
This PR adds a pivot point around which rotations happen to a component.  This makes it easier to rotate a component interactively around an atom, the component's center, or any other point.

Based on that, it adds a `CenterMode` to `StructureComponent` to allow interactively rotating around the current selection or (the default) the structure's center.

This PR also allows the `viewer.rotationGroup` to have a non-unit scale factor to scale the whole scene, for when it is included in a larger three.js scene. It simply extracts the rotation part only in `trackball-controls.ts`. That seemed a lot simpler (and a lot less code change) than introducing a new `viewer.scaleGroup`.

Adds an example to show how it works. There is some complexity to handling selection-changes so it seemed worth creating the fully worked example.

### DIFF
--- a/src/component/structure-component.ts
+++ b/src/component/structure-component.ts
@@ -91,6 +91,8 @@ export interface StructureComponentSignals extends ComponentSignals {
   defaultAssemblyChanged: Signal  // on default assembly change
 }
 
+export type CenterMode = 'selection' | 'component'
+
 /**
  * Component wrapping a {@link Structure} object
  *
@@ -120,6 +122,11 @@ class StructureComponent extends Component {
   dihedralRepresentation: RepresentationElement
 
   measureRepresentations: RepresentationCollection
+
+  // StructureComponent rotates and scales around the selection's
+  // center when centerMode is "selection", otherwise around the whole
+  // structure's center.
+  centerMode: CenterMode = 'selection'
 
   constructor (stage: Stage, readonly structure: Structure, params: Partial<StructureComponentParameters> = {}) {
     super(stage, structure, Object.assign({ name: structure.name }, params))
@@ -201,6 +208,7 @@ class StructureComponent extends Component {
 
       this.rebuildRepresentations()
       this.rebuildTrajectories()
+      this.updateMatrix()
     })
   }
 
@@ -261,8 +269,8 @@ class StructureComponent extends Component {
   }
 
   /**
-   * Overrides {@link Component.updateRepresentationMatrices} 
-   * to also update matrix for measureRepresentations 
+   * Overrides {@link Component.updateRepresentationMatrices}
+   * to also update matrix for measureRepresentations
    */
   updateRepresentationMatrices () {
     super.updateRepresentationMatrices()
@@ -354,8 +362,16 @@ class StructureComponent extends Component {
     return bb
   }
 
-  getCenterUntransformed (sele: string): Vector3 {
-    if (sele && typeof sele === 'string') {
+  setCenterMode(mode: CenterMode) {
+    this.centerMode = mode
+    this.updateMatrix()
+  }
+
+  // Structure-component center can be based on the selection or the whole structure.
+  // Caller can override use of selection-center mode by passing useCenterMode = false.
+  getCenterUntransformed (useCenterMode: boolean = true): Vector3 {
+    const sele = this.selection.string
+    if (useCenterMode && this.centerMode === 'selection' && sele && typeof sele === 'string') {
       return this.structure.atomCenter(new Selection(sele))
     } else {
       return this.structure.center
--- a/examples/scripts/test/pivot.js
+++ b/examples/scripts/test/pivot.js
@@ -1,0 +1,111 @@
+// Handle window resizing
+window.addEventListener('resize', function () {
+  stage.handleResize()
+}, false)
+
+var newDiv = document.getElementById('viewport').appendChild(document.createElement('div'))
+newDiv.setAttribute('style', 'position: absolute; top: 0; left: 20px')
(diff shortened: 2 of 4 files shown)


src/component/structure-component.ts
```diff
@@ -354,8 +362,16 @@ class StructureComponent extends Component {
...
-  getCenterUntransformed (sele: string): Vector3 {
-    if (sele && typeof sele === 'string') {
+  setCenterMode(mode: CenterMode) {
+    this.centerMode = mode
+    this.updateMatrix()
+  }
+
+  // Structure-component center can be based on the selection or the whole structure.
+  // Caller can override use of selection-center mode by passing useCenterMode = false.
+  getCenterUntransformed (useCenterMode: boolean = true): Vector3 {
```
fredludlow, line 372: This removes the option to pass an explicit selection through to this method. `getCenterUntransformed(sele: string)`, replacing it with a boolean arg. 

I think this will break `Component.getCenter` which passes its own args through directly to this method, and is used in a number of examples.

I couldn't see a call to `getCenterUntransformed(true|false)` in this PR, but if you're using it in your code could you replace it with explicit calls to the selection string you want?

src/component/structure-component.ts
```diff
@@ -201,6 +208,7 @@ class StructureComponent extends Component {
 
       this.rebuildRepresentations()
       this.rebuildTrajectories()
+      this.updateMatrix()
```
fredludlow, line 211: This feels odd coupling matrix to selection - should the call instead be to `setPivot` explicitly if centerMode === 'selection' (which then triggers `updateMatrix` from the `setPivot` method itself)?  (I might just have misunderstood what's going on).

fredludlow (changes requested):
Thanks for this @garyo. I have one major request which is that we preserve the function signature `getCenterUntransformed(sele: string)` so that we can continue to pass selections through (though happy to have it handle additional types of arguments if needed).

I spent a while wondering if there was a simpler way to achieve this that avoids `updateMatrix` calls when selection changes (and then the handling you need to do to avoid the structure jumping around when selection changes). I think the issue is that you need to know the pivot point every time the overall matrix gets updated, and changing the pivot point requires the pre-transform you do in the example). i.e. in this model, the pivot is a fundamental part of describing the final position.

I then wondered if it is possible to use the pivot point in the TrackballControl code, so that the `rotateComponent` function actually updates both rotation and translation of the component (according to the position of the pivot) - i.e. the component's `updateMatrix` code would no longer need to worry about the pivot, the pre-transform is done in TrackballControl

The problem with that way is that it makes it hard to do a rotation around an arbitrary point from anywhere other than trackball controls code (you'd have to reimplement your pretransform logic). 

So then I wondered, should the pre-transform from your example go into the `setPivot` method? i.e. should setPivot be written such that the overall matrix does not change? (I've only just thought of that last one, so it's quite possible it's also a bad idea...)

fredludlow:
btw - it builds fine and tests pass etc:
https://fredludlow.com/ngl/pr-909/examples/webapp.html?script=test/pivot

src/component/structure-component.ts
garyo: Good point. I'll fix this so it accepts an optional selection as the first arg (I think it should be optional -- not all callers pass it today), and uses that instead of the current selection when passed in.

src/component/structure-component.ts
garyo: When the selection changes, if the `centerMode` is "selection", the matrix has to be recalculated because the _center_ has changed, not the pivot point. So I don't think it would be clearer to call `setPivot(this.pivot.x, this.pivot.y, this.pivot.z)` rather than `this.updateMatrix()`. How about a comment there explaining why the matrix needs to be updated? And I will do it only when `centerMode === "selection"`.

fredludlow:
Right, I think I understand now:

As you said: When you update selection, the *center* changes. We still want to keep the pivot as a relative offset from the center (the clear case for this is that if it's at 0,0,0 for the whole structure and you select only one chain, you don't want pivot jumping to some offset and negating the nice behaviour of rotating around the center of the selection).

Your example code recalculates `component.transform` such that the final position is unchanged (counteracting the change in center+pivot).

Aside: matrix inversions get a bad rep but I think we're good here - it would be possible to calculate the new transform without the inversion from a combination of old and new centers, but this is shorter (otherwise you need to handle scale too)

I tried moving the selection change code from your pivot example into the selection.signals.stringChanged handler in structure-component.ts, i.e:

```typescript
      this.selection.signals.stringChanged.add(() => {
      this.structureView.setSelection(this.selection)

      this.rebuildRepresentations()
      this.rebuildTrajectories()
      if (this.centerMode === 'selection') {
        // Selection's center point has changed, and in center mode we use
        // the selection center in the matrix (via getCenterUntransformed),
        // so we need to recalculate the matrix.

        // Additionally:
        // When the selection changes, we don't want the molecule to move
        // around. So compute a pre-transform that will take the new matrix,
        // based on the new center point, back to the old matrix. At this
        // point, it's important that the matrix hasn't yet been updated to
        // reflect the new selection's center point.

        const oldMatrix = this.matrix.clone()
      
        this.updateMatrix(true) // get matrix w/ new selection-center (silently, no signals)
      
        // Update pre-transform to make final result same as m0 (old matrix)
        // T' = m0 * m1^-1 * T
        tmpMat4.getInverse(this.matrix)
        this.transform.premultiply(tmpMat4).premultiply(oldMatrix)

        this.updateMatrix()
      }

    })
```

That seems to work for me, though as always I'm probably overlooking *something*...

PR #910, opened 2022-01-26
Update Typescript from 3.8.3 to 4.5.4

garyo:
Trying to keep up with recent tooling. TS 4 is the minimum version for some web components now.

This also pulls in some other more recent dependencies, like eslint and jest.

Typescript 4 is stricter in some cases, so there are a few changes to the code to reflect that. The only significant ones are that typescript 4 correctly doesn't let you delete a non-optional field of an object -- that would leave the object in a broken state. Instead of deleting things like a structureRepresentation's structure, we just dispose of it or set it to an empty object or array. That will have the same effect; the unlinked object can be garbage collected.


package.json
```diff
@@ -70,26 +71,29 @@
...
     "@types/node": "^10.17.9",
     "@types/promise-polyfill": "^6.0.0",
     "@types/signals": "1.0.1",
     "@types/sprintf-js": "^1.1.2",
     "babel-plugin-array-includes": "^2.0.3",
     "babelrc-rollup": "^3.0.0",
     "cpy-cli": "^3.1.0",
-    "jest": "^25.5.1",
+    "eslint": "^8.5.0",
+    "eslint-plugin-import": "^2.25.3",
```
fredludlow, line 83: This brings in eslint (and some plugins) but I'm not sure we have any eslint config or reference to them anywhere else?

package.json
```diff
@@ -70,26 +71,29 @@
...
     "slimerjs": "^1.0.0",
     "source-map-support": "^0.5.6",
     "standard": "^11.0.1",
-    "ts-jest": "25.5.1",
-    "tslib": "^2.3.1",
+    "ts-jest": "^26.5.6",
     "typedoc": "^0.16.11",
-    "typescript": "^3.8.3"
+    "typescript": "^4.5.4",
+    "uglify-js": "^3.4.6"
```
fredludlow, line 96: I think rollup-plugin-terser is doing this job now?

src/stage/mouse-observer.ts
```diff
@@ -228,15 +230,18 @@ class MouseObserver {
...
    * @return {undefined}
    */
-  _onMousewheel(event: WheelEvent & { wheelDelta?: number, wheelDeltaY?: number }) {
+  _onMousewheel(event: Optional<WheelEvent, 'detail'> & { wheelDelta?: number, wheelDeltaY?: number }) {
     if (event.target !== this.domElement || !this.handleScroll) {
       return
     }
     event.preventDefault()
-    this._setKeys(event)
+    this._setKeys({...{detail: 0}, ...event})
```
fredludlow, line 238: Argh, this was a tricky one: The spread operator (and the Object.assign call that it gets transpiled into) both don't copy the prototype, so things like shiftKey modifiers aren't captured. I fiddled a bit with types but given we're dealing with old browsers and suchlike I think it's fair to put `event as any` here?

fredludlow (changes requested):
With a couple of changes this built fine and works for me, I can push my changes if you want to save you recreating them?

package.json
garyo: TBH I'm not sure why I needed that - might have just been required by my own use of it. Fine to omit if things build.

src/stage/mouse-observer.ts
garyo: I'm fine with that.

garyo:
I expect your changes are fine -- no need for me to recreate them. :-) Go ahead & push with those.

package.json
fredludlow: It did used to be a dependency, but I think the config was all for the pre-TS code (My PR 882 removed it and I did make a start at sorting out the linting/formatting workflow but it had turned into a massive time-sink and wasn't very high priority - I might have the appetite to have another go at some point :) ).

fredludlow (approved):

fredludlow:
Thank you @garyo!

PR #916, opened 2022-03-15
Setup Github Actions

fredludlow:
I've added a basic Github Actions config to check out, build, test (should catch typescript issues). 

Will see if I can figure out releasing via pushing a tag or similar in a separate PR

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,0 +1,30 @@
+# This workflow will do a clean installation of node dependencies, cache/restore them, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Node.js CI
+
+on:
+  push:
+  pull_request:
+    branches: [ master ]
+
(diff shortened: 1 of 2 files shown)


ppillot:
Any way to test it?
Otherwise, go for it @fredludlow, it's a very nice addition

fredludlow:
I think you can test by pulling my branch and pushing to your gitlab. There may even be a nice way to do this through the UI but something along the lines of:

```
git remote add fredludlow git@github.com:fredludlow/ngl.git   # Or https://github.com/fredludlow/ngl.git
git fetch fredludlow
git checkout actions   # I think default behaviour is to check out a branch tracking mine
git push origin actions # where origin is your own github repo: git@github.com:ppillot/ngl.git
```
I imagine that would trigger the actions on your ppillot/ngl repo

The nice thing about actions is that for the basic stuff you can just test in your own repo, e.g:
https://github.com/fredludlow/ngl/actions

For publishing on npm it does need to know relevant tokens/secrets, but we can cross that bridge later as publishing from command line works.

Issue #879, opened 2021-09-06
Load PDB directly from AlphaFold Protein Structure Database


nvelden:
With the exciting release of the AlphaFold protein structure database would it be possible to add the option to load PDB files directly from the database using the Uniprot identifier? Access to structures can be achieved using:

https://alphafold.ebi.ac.uk/files/AF-" + pdbid + "-F1-model_v1.pdb

I saw that icn3d viewer has it already integrated. See:
https://github.com/ncbi/icn3d/search?q=alphafold

fredludlow:
Yes, this would be great - it would presumably follow the pattern in https://github.com/nglviewer/ngl/blob/master/src/datasource/rcsb-datasource.ts

PR #920, opened 2022-03-30
Add Alphafold datasource

fredricj:
Add Alphafold datasource based on the RCSB datasource. #879

### DIFF
--- a/src/datasource/alphafold-datasource.ts
+++ b/src/datasource/alphafold-datasource.ts
@@ -1,0 +1,37 @@
+/**
+ * @file Alphafold Datasource
+ * @author Fredric Johansson <fredric@fredricj.se>
+ * @private
+ */
+
+import { Log, DatasourceRegistry } from '../globals'
+import { getProtocol } from '../utils'
+import { getFileInfo } from '../loader/loader-utils'
+import Datasource from './datasource'
+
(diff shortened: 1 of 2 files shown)


fredludlow (approved):
This looks great, thank you!

fredludlow:
Thanks @fredricj, working nicely for me.

Might be worth adding an example script too (happy to push one to your branch and then merge if you like?)

One that follows the color scheme the EBI use on their alphafold page (via @rjdkmr) is below.

Also, more as note to self - I really like this color scheme for alphafold, and I think there's a benefit in consistency across software and viewers so maybe as a future change extending defaultRepresentation to take options like: 'protein'|'alphafold'|'ligand' 

```
stage.loadFile("alphafold://P24941").then(function(c) { 
  c.addRepresentation("cartoon", {
    colorDomain: [0, 49.999, 50, 69.999, 70, 89.999, 90, 100],
    colorScale: ['#ff7d45', '#ff7d45', '#ffdb13', '#ffdb13', '#65cbf3', '#65cbf3', '#0053d6', '#0053d6'],
    colorScheme: 'bfactor'
  })
})

fredludlow:
Thanks again @fredricj - I've added an example as test/alphafold.js

PR #922, opened 2022-04-12
Use stable angle method (atan2) to avoid precision problems in dihedr…

fredludlow:
Very quick/minor fix for a case I had when measuring dihedrals at very close to 180 degrees. Floating point rounding errors meant we were trying to calculate `Math.acos(-1.000001)` or similar which produces `NaN`, this instead uses the `v3angle` method (`Math.atan2` underneath) which doesn't suffer these kind of precision/domain issues.


fredludlow:
Going to just merge this, please shout in the unlikely event it causes problems

PR #926, opened 2022-04-22
Enable three.js treeshaking with three-minifier-rollup

milot-mirdita:
This reduces the bundle size by a few hundred kb. I added the bundle both as ESM and UMD.

It would be nice if ngl itself could be treeshaken, e.g., the parsers and representations are quite large.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "ngl",
-      "version": "2.0.0-dev.39",
+      "version": "2.0.0-dev.40",
       "license": "MIT",
       "dependencies": {
         "chroma-js": "^1.3.7",
@@ -31,6 +31,7 @@
         "@types/promise-polyfill": "^6.0.0",
         "@types/signals": "1.0.1",
         "@types/sprintf-js": "^1.1.2",
+        "@yushijinhun/three-minifier-rollup": "^0.3.1",
         "babel-plugin-array-includes": "^2.0.3",
         "babelrc-rollup": "^3.0.0",
         "cpy-cli": "^3.1.0",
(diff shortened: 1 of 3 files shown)


fredludlow:
Thanks for this @milot-mirdita, it looks useful - I'll take a proper look next week.

Two quick points:
* This renames ngl.js to ngl.bundle.umd.js and adds ngl.bundle.esm.js. I worry that people may just be using dist/ngl.js (rather than packaging mechanisms) and this will break workflows. 
* Also, what's the use-case for ngl.bundle.esm.js, as opposed to using ngl.esm.js as specified in package.json's module attribute?

As for tree-shaking NGL itself, I guess you'd need an entry point that exposed a minimal core and then you'd import those parsers, representations, writers etc that you wanted to use? Would be a nice addition.

milot-mirdita:
>     * This renames ngl.js to ngl.bundle.umd.js and adds ngl.bundle.esm.js. I worry that people may just be using dist/ngl.js (rather than packaging mechanisms) and this will break workflows.
>     * Also, what's the use-case for ngl.bundle.esm.js, as opposed to using ngl.esm.js as specified in package.json's module attribute?

I wanted to be consistent and build both an umd and an esm version of the bundle with the tree-shaken three.js. I can leave the umd version with the old name and just add an esm version with a new name, or alternatively drop the esm version. But I think that esm is a prerequisite to tree-shaking (among others) if i understood it correctly. 

> As for tree-shaking NGL itself, I guess you'd need an entry point that exposed a minimal core and then you'd import those parsers, representations, writers etc that you wanted to use? Would be a nice addition.

Something like that would be great. NGL is currently the largest dependency, any byte helps :)

fredludlow:
Right, getting back to this now.

Can you give me an example of how that bundle.esm would actually get used? Prior to this PR, the `rollup.config.dist.js` produces:

1. `dist/ngl.umd.js`  - Three not included. This is for importing via classic node `var NGL = require("ngl")` syntax, it's probably not used much tbh
2. `dist/ngl.esm.js` - Three not included. If you specify NGL as a dependency then npm/yarn/etc should install it to node_modules together with the dependencies we specify. Your build tool (rollup/webpack whatever) should include the correct version of those dependencies. In this case, you would want to include the three-minifier-rollup plugin in your downstream project's build configuration
3. `dist/ngl.js` - A UMD format "everything" build so you can just drop in a `<script src="/path/to/that/file">`  (this is what you've renamed `dist/ngl.bundle.umd.js`

I'm in favour of saving 300Kb in for option 3 (which is half your PR). I'm not so keen on the rename as it'll probably break stuff for other people using this.

I'm still not sure why `dist/ngl.bundle.esm.js` is necessary - it's not referenced in package.json so won't be picked up unless you manually point at it, but wouldn't you get the same benefit importing NGL as a module from your app and using three-minifier-rollup in your own build config?

I guess if you're using import/export [directly in the browser](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#browser_support) you might want to split your app into a slowly-changing "NGL+immediate deps" ES6 module plus a faster-changing "my-app" module to make the most of browser caching - but this is a very specific point to draw the line and seems out of scope to have in the main NGL distribution?

Obviously if I'm completely missing the point please do tell me!

milot-mirdita:
I mostly changed it for consistency, without much deeper thought.

I did later try to use ngl as a `<script type="module">` in our [CASP15](https://casp15.colabfold.com) visualization but quickly gave up and used the umd version instead.

fredludlow (approved):
Thanks for the PR and for requested changes, looks good.

Issue #865, opened 2021-05-05
Deprecated `update` package


omaraltayyan:
Hi, 

Can we deprecated or remove the `update` package we use? since doing an `npm audit` over the repo after installing ngl shows tons of security vulnerabilities, I haven't seen this package before but it says that it's some sort of release helper? should it be moved to dev dependencies instead?

fredludlow:
Yes, we should squash all those audit warnings

omaraltayyan:
@fredludlow if this helps, changing the imports of three.js of `Geometry` and `Face3` to 

```
import {Geometry, Face3} from 'three/examples/jsm/deprecated/Geometry';
```

Makes the code compatible with minimal changes

tom-cresset:
Would be great to see `three` updated to appease audit, etc.

contramundum53:
Any updates on this issue?

papillot:
@contramundum53  No update unfortunately.
I am not sure to understand the implications of updating Three.js. Apparently there has been changes to the web gl support, and other changes to the Geometries and some API changes.
There was a PR made about 1 year ago which was in part doing that: https://github.com/nglviewer/ngl/pull/951 but had been closed. It might be a base for whomever feels ready for this task.

ppillot:
So, I gave this a try. I have been able to update versions of three.js up to r154.
There is a major change wrt to lighting in r155 which leads to such results:
<img width="367" alt="image" src="https://github.com/nglviewer/ngl/assets/3930632/250c4c3e-ce71-4e74-a57f-67cbe390693c">
It's possible to increase the ambient light, but then all colors remain flat
<img width="356" alt="image" src="https://github.com/nglviewer/ngl/assets/3930632/8a04c9fd-fd4c-4f31-8566-81e4a78c2181">
I will create a PR with the changes up to r154 and give some time to test how it fares.

ppillot:
I've been able to upgrade to the latest version by using the deprecated `.useLegacyLights` property.
It seems that there are several properties in the lights that should be set (like the distance and the decay) on top of the change of the internal intensity (which was corrected by a factor set to PI internally).
See some details here: https://discourse.threejs.org/t/updates-to-lighting-in-three-js-r155/53733

PR #929, opened 2022-05-12
Housekeeping

fredludlow:
This updates some dependencies and squashes a few deprecation/audit warnings.

The one I haven't addressed is updating three (which has now deprecated the old `Geometry` type as mentioned in #865). It's currently not causing any issues for me having an older version of three.js but getting too far behind may be painful in the long run. Do others have any views on this?


luxaritas:
I'm getting a dependabot security warning for three < 0.125: https://nvd.nist.gov/vuln/detail/CVE-2020-28496

PR #944, opened 2022-11-10
Support V3000 in sdf-parser.ts

pablowatson:
We noticed that V3000 mol/sd files are not supported in the sdf-parser. To fix that, this fork contains some changes to handle both V2000 and V3000 mol/sd files. I also added some rudimentary tests.


ppillot:
Thanks for this welcome addition!
I'll take the time to review the proposed changes in more details, but please note that the `dist` directory should not be submitted in a PR as these files are autogenerated when a new release is made and they add a sensible amount of data to the commits.

ppillot (changes requested):
Sorry about that, I realize that I forgot to post the review I made a few weeks ago...

Anyway, I've decided to try to follow your steps and implement V3000 parsing.
Doing so I've noticed that the SDF format allows mixing V2000 and V3000 CTABs in the same SDF file, which requires to be able to change the parser within the same file.

Also, I've stumbled on examples of V3000 files (e.g. in RDKit test files and in Biovia documentation)  where the `V3000` format token is not  at the 35th column of the 4th line.

Lastly, the number of lines in the atom block or the bond block can't be guaranteed based on the COUNT line, as it is possible to write on multiple lines using the `-` token at the end of a line. This is probably a minor concern as most of the atom fields are not relevant to 3D structures.
I'll submit a joint PR hopefully shortly.

pablowatson:
Looking forward to the joint PR. Thanks for taking the time to look at this.

ppillot:
@pablowatson you may have a try with PR #955 
I've reused the test files in their entirety. The code for the parser has been heavily modified. It would be much appreciated if you could have a look at it based on your experience!

pablowatson:
@ppillot Sorry for not getting back to you sooner. I will take a look at the code this week. Thanks.

ppillot:
Superseded by PR #955

PR #951, opened 2022-12-13
Sonarcloud

sjackson-ccdc:


ppillot (changes requested):
Thank you for your contribution!

This PR upgrades the dependencies concerning THREE.js.
But it also contains several unrelated changes:
- changes in rollup build process by removing buble
- changes in PLY parser, with changes to the PLY parser tests (I don't know how to analyze these)
- adds Sonarcube GitHub workflow depending on secret keys that are not part of this upstream repository configuration (I am not sure of the implications there?)
- many files with empty changes are part of this PR:
  - test/parser/test-cif-parser.spec.ts
  - test/parser/test-cube-parser.spec.ts
  - test/parser/test-pdb-parser.spec.ts
  - test/selection/tests-selection.spec.ts
  - test/structure/tests-structure-view.spec.ts
  - tsconfig.json
  - tslint.json

To move forward with the changes in this PR, would it be possible to split it? I would suggest to create a distinct PR for the changes related with the THREE.js upgrade.
Please review your changes to avoid committing changes in indentation (in several cases a space is inserted at the start of a line without introducing any other change).

Regarding the PLY parser, the changes look non trivial. Is it also an upgrade from third party code? Is it related with the changes made to THREE.js?
Could you also explain why the tests had to be changed. My understanding is that the input file is still the same, so I would expect the number of vertices/faces/normals etc... to be conserved, but maybe the previous version was not working properly? Do you have any insights on this?

A distinct PR could be made regarding the removal of buble. I think it has been discussed in the past, but this should be investigated separately.

Finally, regarding sonarcloud, I don't know if this should be part of this repo. This proposal should be discussed separately and I would appreciate to have your analysis on what this would entail to the main NGL project and anyone who forks it on GitHub

PR #976, opened 2023-03-23
Actions housekeeping

fredludlow:
This updates the node versions for CI (removes v12, adds v18) and updates some of the dependent actions to latest versions which squashes a lot of warnings about deprecated calls to save-state and similar

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [16.x, 18.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
```
ppillot, line 18: node 14 will actually reach EOL next month (April 2023). It might be worth demoting it now

ppillot (approved):
Looks good.
Consider removing Node 14.x as well (EOL in April 2023)

.github/workflows/node.js.yml
fredludlow: Good point

PR #977, opened 2023-03-23
Infer bond options

fredludlow:
For some PDB format files, I am seeing extra bonds drawn between atoms when they are within what would normally be considered covalent range but not actually bonded (I'm trying to find some good CSD examples I can share and will update with an appropriate test case). The current rules for detecting bonds use sum of some preset covalent radii and a threshold, but some of these structures legitimately fall within this threshold, though there is no bond.

This PR adds an `inferBonds` parameter when loading a structure, currently only consumed by the PDB parser class. It's behaviour is:

* `all` - Default, existing behaviour - creates bonds between any atoms within in a residue according to distance rules
* `auto` - For a het group residue (and only het groups), if any bonds are specified (by CONECT records) then only use the CONECT records, otherwise behave as before
* `none` - Don't infer any bonds at all based on distance

It also makes `structureBuilder` a declared member of `StructureParser` (it is initiated in it's constructor already) and fixes up a few minor type errors that result from having the type information.

As I said, I'll add a test but wanted to push what I had so far in case anyone has a view on doing it differently.

### DIFF
--- a/src/loader/parser-loader.ts
+++ b/src/loader/parser-loader.ts
@@ -5,6 +5,7 @@
  */
 
 import { ParserRegistry } from '../globals'
+import type { InferBondsOptions } from '../structure/structure-utils'
 import Loader from './loader'
 import { LoaderParameters, LoaderInput } from './loader-utils'
 
@@ -18,6 +19,7 @@ export interface ParserParams {
   delimiter?: string
   comment?: string
   columnNames?: string
+  inferBonds?: InferBondsOptions
 }
 
 /**
@@ -37,6 +39,7 @@ class ParserLoader extends Loader {
       delimiter: params.delimiter,
       comment: params.comment,
       columnNames: params.columnNames,
+      inferBonds: params.inferBonds,
       name: this.parameters.name,
       path: this.parameters.path
     }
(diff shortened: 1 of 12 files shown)


fredludlow:
Yes, we should put the autoformatting back in!

pschmidtke:
@fredludlow do you have a public domain structure (not CSD, but PDB) where this happens ? I observed that issue on some in-house structures sometimes. 

How would explicit definitions of CONECT records influence this option here?

fredludlow:
I don't have a public domain PDB structure unfortunately. I've dug out a couple of CSD examples where the NGL bonding threshold says there is a bond but the CSD structure says there isn't (at least to a first approximation of where to draw the bonds, some of the examples are interesting organometallics). I'm not sure how to do the equivalent search on COD..?

If you have anything public domain like this that would be great, if not then I can snip a bit of a CSD ligand, or even, given it's just to enumerate some test cases, take butane and stretch one of the terminal bonds into this ambiguous region

ppillot:
@fredludlow what is the status for this branch?

fredludlow:
We've been using it in our internal build since the PR was created, no issues reported locally - it could do with a simple test case (I'll just do ethane with varying bond distances to exercise the different options)

ppillot:
Ok, let's merge it then

ppillot (approved):

PR #979, opened 2023-04-05
Fix calculate chain names

papillot:
This PR adresses a bug in the modelStore consistency when the `calculateChainnames()` function is called.
When the initial number of (unnamed) chains was greater than the number of chains calculated by the method, the chainCount in the modelStore was incorrect which could lead to issues in the execution of the `model.eachChain` method.
A symptom of that was duplicated records in the PDB output from the `PdbWriter` class.

The coding issue was a missing reinitialization of the chains count in the modelStore before creating the new chains. There was a post correction of this number by a decrement but this was not guaranteed to work in every case (in fact it was working on the assumption that a single chain was initially created).

The PR contains a test file witch replicates the non-standard PDB file that revealed that bug.


ppillot:
Do not merge yet: the list files should not have been in the PR

ppillot:
Offending commit has been removed. Ready for review/merge

fredludlow (approved):
Again, looks great, thanks!

PR #984, opened 2023-04-25
Fix pdbwriter model format

papillot:
This PR fixes 3 issues with the MODEL records in the PDB writer:
- It was initially detected while using the AutoDock Tools for preparing a receptor file using a PDB file generated via NGL.
Here is the permalink to the code line that is extracting the MODEL number value: https://github.com/Pymol-Scripts/Pymol-script-repo/blob/11247c46fe05c6e22f63d66b2f99526b71253e2c/modules/ADT/MolKit/pdbParser.py#L140
It is expected that the model index is displayed in the columns ranges 11-14 as per the PDB specs.

Note also that as part of the PDB specs for this record, the MODEL record shall only be included when there is more than one model in the structure.

Finally there was a bug in the numbering when there is more than one model, where the MODEL ID got incremented twice

PDB specs: https://www.wwpdb.org/documentation/file-format-content/format33/sect9.html#MODEL


fredludlow (commented):
Looks great, thank you

fredludlow (approved):

PR #988, opened 2023-06-08
Fix/#594

HHongSeungWoo:
First of all, I want to thank you for making such a great library.

I'm using ngl with React.
When I install ngl with npm and build with swc, I get the error
 ```SyntaxError: function statement requires a name```

The problem is that the result of the getAllowedContours function is inlined and the name removed.
After this, in the makeWorkerString function, the error occurs as a result of ```(function () { return [...] }).toString()```

So I removed getAllowedContours from __deps and inlined it.

### DIFF
--- a/src/surface/marching-cubes.ts
+++ b/src/surface/marching-cubes.ts
@@ -304,42 +304,6 @@ function getTriTable (): Int32Array {
   ])
 }
 
-// Triangles are constructed between points on cube edges.
-// allowedContours[edge1][edge1] indicates which lines from a given
-// triangle should be shown in line mode.
-
-// Values are bitmasks:
-// In loop over cubes we keep another bitmask indicating whether our current
-// cell is the first x-value (1),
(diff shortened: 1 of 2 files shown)


fredludlow (approved):

fredludlow:
Available in 2.1.1 https://www.npmjs.com/package/ngl

HHongSeungWoo:
@fredludlow 

webpack and swc both use or reference terser for minification and optimization. The ```getAllowedContours``` function is a pure function that returns a constant value without any arguments, and it is called only once in the code. Therefore, the single invocation gets inlined by directly assigning the return value, and ```getAllowedContours``` is no longer used.

Since the function is being passed in the ```__deps``` without being called anymore, during the process of minification and mangling, its name is removed and it gets inlined.

There are a few possible solutions to this issue. One option is to refactor the code. Another option is to configure the pure_funcs or keepfnname options in SWC or webpack. However, the simplest solution in this case is to assign the value directly.

PR #992, opened 2023-06-27
added scroll for focusScroll in pymol mouse controls

chmnk:
Both scroll and drag-shift-right do focusScroll action in pymol, but scroll is more intuitive, because this makes controls mouse-only.

### DIFF
--- a/src/controls/mouse-actions.ts
+++ b/src/controls/mouse-actions.ts
@@ -246,6 +246,7 @@ export const MouseActionPresets = {
     [ 'drag-left', MouseActions.rotateDrag ],
     [ 'drag-middle', MouseActions.panDrag ],
     [ 'drag-right', MouseActions.zoomDrag ],
+    [ 'scroll', MouseActions.focusScroll ],
     [ 'drag-shift-right', MouseActions.focusScroll ],
 
     [ 'clickPick-ctrl+shift-middle', MouseActions.movePick ],


fredludlow:
I'm not familiar with pymol controls but this seems sensible. @ppillot ?

papillot:
I don't use Pymol either. In the version I have, the mouse wheel seems to only move the front clipping pane. Zooming is done through right click or wheel + Ctrl + Shift
image
I agree that scroll to zoom seems more intuitive, but I am not sure it's the default specification for Pymol?

chmnk:
>  Zooming is done through right click or wheel + Ctrl + Shift

@papillot In your figure the **editing** mode is shown. I think the default pymol's mode is **viewing**, and the current pymol configuration in ngl seems to represent the **viewing** mode.

>  I agree that scroll to zoom seems more intuitive

No, I am not talking about zooming, I meant clipping. Zooming in pymol's viewing mode is the `drag_right` and that's nice and clean, and is already implemented in ngl.

> In the version I have, the mouse wheel seems to only move the front clipping plane.

Following this reference https://pymol.org/dokuwiki/doku.php?id=mouse:clipping, in the viewing mode, scroll expands or shrinks the area between the clipping planes. That area is called "slab". 
As it is better explained here https://pymol.sourceforge.net/newman/user/S0200start.html, `shift+drag_right` has more flexible behaviour that depends on the direction of dragging (up-down / left-right for moving front/back clipping planes instead of moving them as whole).

What I experienced in ngl with `focusScroll` reminded me the simple `scroll` action in pymol, which is the motivation of this request.

PS Actually when I started this, I didn't know the details of how does shift+drag_right work in pymol (I thought it's the same as scroll)

ppillot (approved):
Thanks for the addition

PR #993, opened 2023-07-01
Remove support for IE11

HHongSeungWoo:
IE11 is deprecated.
This change will save a lot of bundle size.


papillot:
Thanks for looking into this!
Do you by any chance have any metrics on the before/after regarding the bundle size?

Also, it looks like a good idea to remove the polyfills as well. I've been through the list and noted that they polyfill IE only missing feature (in other words these features were implemented circa 2015 or before) 

Also, what is the consequence of removing buble? Is there any kind of transpilation yet? What will be the final code compatibility target?

HHongSeungWoo:
@papillot 

## Compare bundle sizes
```
egl.esm.js: 760kb -> 677kb
ngl.js: 1.2mb -> 1.1mb
ngl.umd.js: 765kb -> 682kb
```

The bubble compiler seems to exist to support ES5 syntax in IE11. Browsers that don't support ES5 are currently deprecated, so it should provide the same compatibility as before.

HHongSeungWoo:
@papillot Could you please review this PR?

ppillot:
@HHongSeungWoo I'll make my best to finalize this in the upcoming days.
Would you mind removing the build commits from your PR (`git rebase -i` and remove the two build commits) ? We'll update dist for the next released version and in the mean time we try to avoid commits with build files in them. Thank-you for your contribution.

ppillot (approved):
Thanks for the improvements.
Will do a release with these changes soon.

Issue #999, opened 2023-11-13
NGL error displaying some mmCIF files


brucelab:
Lately I've noticed that many mmCIF files do not display in NGL and give an error: 'TypeError: s.name.trim is not a function.'  For example the file 32c2.cif is one that causes the error and is not displayed on the NGL web application at https://nglviewer.org/ngl/.

On close inspection, I have found that mmCIF files that contain _chem_comp_atom, entries seem to cause the problem.  When I delete that section of an mmCIF file, or even just rename the _chem_comp_atom. entries '_chem_comp_atom2.' the file is displayed in NGL without a problem.

Do you know about this issue?  Is it possible to enable mmCIF files with the _chem_comp_atom. entries to be displayed?

hainm:
We got the issue in `nglview` too (not able to show .CIF file): https://github.com/nglviewer/nglview/issues/1082#issuecomment-1811176285

ppillot:
@brucelab @hainm For the record this was due to an assumption that when this property is defined, the cif file is not a mmCif file and most probably the kind of cif file that can be dowloaded from the PDB ligand repository.
As you've noticed this assumption can be false in some mmCIF files. In that case, this cif loop contains the list of amino acids which causes an error in the code execution.
The fix done is to check if the `struct` property is set, to switch to the proper parsing routine.

Issue #950, opened 2022-12-11
NGL shows white color for this cif


hainm:
Dear all, how are you doing?

One of `nglview`'s users gets issue with color of his cif file (I am attaching here).

<img width="1344" alt="image" src="https://user-images.githubusercontent.com/4451957/206934976-97a70880-80d0-4e84-a1da-0228f0d1292f.png">

[V4GaSe8.cif.txt](https://github.com/nglviewer/ngl/files/10203546/V4GaSe8.cif.txt)

Do you have any suggestion? Thanks
PS: nglview issue is here: https://github.com/nglviewer/nglview/issues/1039

ppillot:
White is the default element color when the element is unrecognized.
I have had a look at the parser code for core CIF files (i.e. not mmCIF files) and it appears that the element name is taken from `_atom_site_type_symbol` without processing.
In the case of the file you provided, that means that `V3.25+` will be used as element name in lieu of `V`, and it won't be recognized as a known element further down.
The fix shall be easy to make and we may as well infer the formal charge value (though it might be preferable to get it from the `_atom_type` loop)

hainm:
Thanks @ppillot

ppillot:
Fixed:
<img width="804" alt="image" src="https://github.com/nglviewer/ngl/assets/3930632/2c64cd67-7300-4439-8122-e1d417b843b1">

PR #1001, opened 2023-11-19
#999 cif parsing

ppillot:
This PR fixes 2 bugs in relation with CIF files parsing:
- #999 chem_comp can be set for macromolecular cif files and contain vectors of entities. In the case of mmCIF files the block of code that parses files containing only one entity (such as PDB het cif files) must not be reached.
- #950 core cif files often define charges as part of the atom type symbol. In that case, the element symbol must be stripped of the charge.

### DIFF
--- a/src/parser/cif-parser.ts
+++ b/src/parser/cif-parser.ts
@@ -25,6 +25,7 @@ const reWhitespace = /\s+/
 const reQuotedWhitespace = /'((?:(?!'\s).)*)'|"((?:(?!"\s).)*)"|(\S+)/g
 const reDoubleQuote = /"/g
 const reTrimQuotes = /^['"]+|['"]+$/g
+const reAtomSymbol = /^\D{1,2}/ // atom symbol in atom_site_label
 
 interface Cif {[k: string]: any}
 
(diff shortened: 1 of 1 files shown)


hainm:
@ppillot  are you able to reproduce the bug without your change and see if the change actually fixes the issue?

hainm:
Also, do you have any idea why `nglview` fails to show `3pqr` recently, althought it has been working well for years (because we have been showing it as example in our `nglview` website)

image

Here is the code in `nglview`: https://github.com/nglviewer/nglview/blob/d9e7740100b586b1e73cc51d247d16b6dc8d0914/nglview/adaptor.py#L123-L125

fredludlow (approved):
LGTM - Tested locally. Both the 3pqr.cif file (downloaded from http://www.rcsb.org/pdb/files/3pqr.cif) and the one posted under #950 were previously reproducibly broken, but are now working with this patch.

hainm:
Thank you. Hopefully there will be a release soon.

ppillot:
@hainm release done and available on npm

hainm:
Beautiful. Thank you very much.

Issue #1010, opened 2024-01-06
Issue(s) with release 2.2.1 (e.g: custom color doesn't work)


hainm:
Hi,

# custom color
While I am testing NGL 2.2.1 for `nglview` (https://github.com/nglviewer/nglview/pull/1086) , I find that the custom color doesn't work anymore

<img width="1496" alt="image" src="https://github.com/nglviewer/ngl/assets/4451957/b6d0612b-17e2-4545-b557-3287663c4d34">

ppillot:
Thanks for reporting this. I've noticed the same issue in a separate branch and did not realize it was still in production.

hainm:
Thanks @ppillot. I look forward for the fix.

ppillot:
@hainm fix has been published to npm as v2.2.2

hainm:
@ppillot I confirm the fix does work. Thank you so much, Paul.

PR #1011, opened 2024-01-08
#1010 Colormaker-constructor

ppillot:
This PR fixes issue #1010 where the syntax for registering a ColorMaker does not generate valid JS.

After bissection, I've found out that this issue has been introduced with the recent changes to the build process (PR #993). It's possible that the transpilation + bundling results in invalid JS code (?).
The fix was already done in the update-three,js ongoing branch, and had been cherry-picked from there.

### DIFF
--- a/src/color/colormaker-registry.ts
+++ b/src/color/colormaker-registry.ts
@@ -201,15 +201,13 @@ class ColormakerRegistry {
     delete this.userSchemes[ id ]
   }
 
-  _createScheme (constructor: any) {
-    const _Colormaker = function (this: any, params: ColormakerParameters) {
-      Colormaker.call(this, params)
-      constructor.call(this, params)
+  _createScheme (constructor: any): typeof Colormaker {
+    class _Colormaker extends Colormaker {
+      constructor (params: ColormakerParameters) {
+        super(params)
+        constructor.call(this, params)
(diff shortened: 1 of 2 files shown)


fredludlow (approved):
That is a little strange, and surprising why the existing classical (explicit prototype) method fails, but switching to the class syntax reads more clearly anyway.

I take it this works with:

a. Regular built-in colorschemes (uniform etc)
b. Colorschemes added by a user which implement atomColor etc
c. Selection colorschemes?

If so then all looks good to me.

ppillot:
@fredludlow I've made some further tests and, noticed that one of my projects, it breaks the Typescrip checks. If I ignore the checks, the code works well though.

image

I'll investigate further.

ppillot:
Issue resolved: it was a missing typescript definition for the argument of the `addScheme` method.
I've checked on a personal project and in every "Color" example of the NGL webapp.

ppillot:
> That is a little strange, and surprising why the existing classical (explicit prototype) method fails, but switching to the class syntax reads more clearly anyway.

I am wondering if that's not an issue with terser, which can turn `function` declarations in more compact arrow functions. It's supposed to avoid doing so when there a `this` in the function body...

PR #1055, opened 2024-11-01
New shape: Icosahedron

tom-daff-cresset:
Adds a new shape so that icosahedra can be created directly.

We use icosahedra for the display of field points around a molecule.


fredludlow (approved):

fredludlow:
Thanks! And sorry for delay reviewing

PR #1063, opened 2026-07-23
Fix impostor clip zoom

Aenardhil:
I've needed to show molecule in very high zoom so when I changed clipDist to 0 and zoomed, the bond was disapearing halfway to the end of the bond.

### DIFF
--- a/examples/scripts/test/clip-dist-zoom.js
+++ b/examples/scripts/test/clip-dist-zoom.js
@@ -11,6 +11,10 @@
 //
 // How to test: run the scene, then scroll-zoom hard onto a single bond.
 // Press 'c' to toggle clipDist between 0 (zoom freely) and 10 (default floor).
+// Press 'o' to toggle perspective / orthographic and check both camera modes.
+// (In orthographic mode zooming magnifies rather than moving the camera in,
+// and gl_Position.w is always 1, so the near-clipping the fix addresses does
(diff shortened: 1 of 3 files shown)


fredludlow:
Thanks, I don't have a huge amount to work on NGL now but this does look sensible. Does it behave correctly in orthographic mode too as well as perspective?

Aenardhil:
Yes — it's correct in orthographic too. In ortho gl_Position.w is always 1, so the clip test is just -1 ≤ z ≤ 1 and the per-corner w never shrinks toward zero the way it does in perspective — so the vanishing the fix addresses can't occur there in the first place. Both the old 0.99 and the new 0.0 sit safely inside the range, so ortho behavior is unchanged. I added an o key to the example scene to toggle perspective/orthographic so it's easy to confirm both modes.

I've updated the PR to have an example that let you validate functionality pressing "o" key that toggles between orthographic and prespective view. And I don't see any visually problem when the camera is really close.

Aenardhil:
@fredludlow Gentle ping on this one — no rush if you're still short on time. I answered the orthographic question above and added an `o` key to the example scene so both projection modes can be checked quickly. Happy to rebase or split anything up if that makes it easier to review. Thanks!

fredludlow (approved):

fredludlow:
Apologies for the delay, I was away most of August

